PR #3, opened 2021-05-26
Enable go module support for the project

sadmansakib:
through this series of commits i have enabled go module support by following official [docs](https://blog.golang.org/using-go-modules). All README files are subjected to further changes


lggomez:
Thanks for making this initial work

I already made the request to activate travis on this repo and see if we can bring it up on this PR, altough we're also thinking about using github actions travis already provides a simple and known way to start a test matrix so there's that

go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/golang-jwt/jwt
+
+go 1.16
```
oxisto, line 3: This should probably be set to 1.14 to be consistent with the matrix build, possibly even lower

rsa_pss.go
```diff
@@ -15,7 +15,7 @@ type SigningMethodRSAPSS struct {
 	// VerifyOptions is optional. If set overrides Options for rsa.VerifyPPS.
 	// Used to accept tokens signed with rsa.PSSSaltLengthAuto, what doesn't follow
 	// https://tools.ietf.org/html/rfc7518#section-3.5 but was used previously.
-	// See https://github.com/dgrijalva/jwt-go/issues/285#issuecomment-437451244 for details.
+	// See https://github.com/golang-jwt/jwt/issues/285#issuecomment-437451244 for details.
```
oxisto, line 18: Please refrain from just blindly doing a `s/dgrijalva/jwt-go/golang-jwt/jwt`. Until we know if we can transfer the old repositories, links to issues still need to refer to the old repo.

rsa_pss_test.go
```diff
@@ -106,7 +106,7 @@ func TestRSAPSSSaltLengthCompatibility(t *testing.T) {
 		},
 	}
 
-	// Behaves as before https://github.com/dgrijalva/jwt-go/issues/285 fix.
+	// Behaves as before https://github.com/golang-jwt/jwt/issues/285 fix.
```
oxisto, line 109: see above

oxisto (changes requested):

oxisto:
There is an additional issue wrt to the go modules path. If we use `github.com/golang-jwt/jwt` but follow the original release cycle, i.e. `v3.x`, then this will break things. So we either need to use `github.com/golang-jwt/jwt` or re-start the release cycle from `v1.0.0`

go.mod
sadmansakib: done

rsa_pss.go
sadmansakib: sorry i blindly changed those without knowing.

sadmansakib:
> There is an additional issue wrt to the go modules path. If we use `github.com/golang-jwt/jwt` but follow the original release cycle, i.e. `v3.x`, then this will break things. So we either need to use `github.com/golang-jwt/jwt/v3` or re-start the release cycle from `v1.0.0`

Is there any discussion regarding whether this fork will start from v1.0.0 release cycle or inherit v3.x release cycle from dgrijalva/jwt-go

mfridman:
> > There is an additional issue wrt to the go modules path. If we use `github.com/golang-jwt/jwt` but follow the original release cycle, i.e. `v3.x`, then this will break things. So we either need to use `github.com/golang-jwt/jwt/v3` or re-start the release cycle from `v1.0.0`
> 
> Is there any discussion regarding whether this fork will start from v1.0.0 release cycle or inherit v3.x release cycle from dgrijalva/jwt-go

I created another issue #5 here to track adding module support. But as previously mentioned, this all depends on the preference of the original maintainer ([see comment here](https://github.com/dgrijalva/jwt-go/issues/462#issuecomment-846423835)).

After what happened with go-chi (https://github.com/go-chi/chi/issues/561), going from pre-modules v4 to module support with v1 (reasoning explained here https://github.com/go-chi/chi/pull/560) and ultimately settling on /v5 .. I'm a bit reluctant to merge this until we know the preference of the upstream repo.

oxisto:
@sadmansakib: Could you please restore the `.travis.yml` file? thank you! otherwise, LGTM, once all the maintainers agreed on the release cycle.

sadmansakib:
> @sadmansakib: Could you please restore the `.travis.yml` file? thank you! otherwise, LGTM, once all the maintainers agreed on the release cycle.

I have reverted the commit and resolved the conflicts

.travis.yml
```diff
@@ -5,8 +5,7 @@ script:
     - go test -v ./...
 
 go:
-  - 1.7
-  - 1.8
-  - 1.9
-  - 1.10
+  - 1.14
```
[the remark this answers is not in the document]
sadmansakib: i see my bad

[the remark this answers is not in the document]
sadmansakib: @oxisto is it okay now?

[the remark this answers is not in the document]
oxisto: Looks good. You also removed `tip` as well, but that's ok, that was actually accidentally left over, so everything looks good.

oxisto (approved):

Waterdrips (approved):

mfridman (approved):
lgtm. Thank you for your contribution.

PR #12, opened 2021-05-28
Fix issue with MapClaims VerifyAudience []string

Waterdrips:
There was an issue in MapClaims's VerifyAudiance where a []string (which
is valid in the spec) would return true (claim is found, or nil) when required
was not set.
It now checks interface types correctly and has tests written

Fixes: #6 
Signed-off-by: Alistair Hey <alistair@heyal.co.uk>


oxisto:
Ah perfect, a fix without touching the claims struct 👍

claims.go
```diff
@@ -90,15 +90,24 @@ func (c *StandardClaims) VerifyNotBefore(cmp int64, req bool) bool {
...
 		return !required
 	}
-	if subtle.ConstantTimeCompare([]byte(aud), []byte(cmp)) != 0 {
-		return true
-	} else {
-		return false
+	var stringClaims string
+	for _, a := range aud {
+		if subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0 {
+			return true
```
oxisto, line 100: Just a minor thing, but this probably breaks the constant time checking, because it immediately returns once one audience was correct vs. checking all audience whether they are correct or not in constant time. 

Not sure, if constant time is really needed for the audience, through.

map_claims_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+
+		// interface{}
+		{ Name: "Empty interface{} Aud without match not required", MapClaims: MapClaims{"aud": nilInterface}, Expected: true, Required: false, Comparison: "example.com"},
+
+	}
+
+
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
```
oxisto, line 60: This breaks compatibility with Go < 1.7. But I suggest we can safely drop support for those versions :)

Waterdrips:
Have we decided what go versions to support and if we are using github actions OR travis ?

claims.go
Waterdrips: good spot, how about checking everything and using a var to store if any matched?

Waterdrips: changed to use a var and always check all items in the slice

oxisto:
> Have we decided what go versions to support and if we are using github actions OR travis ?

Basically I have enabled GH actions for those versions we want to support (> 1.11). Travis has 1.7-1.10 currently. My idea as then to gradually remove versions from it until Travis can be removed completely.

oxisto (approved):
LGTM

lggomez (approved):
LGTM

PR #14, opened 2021-05-28
Providing (almost) full test matrix in GitHub actions

oxisto:
Starting with Go 1.7, since this is the lowest version still working, once we merge in #12. We can then gradually move upwards with new PRs.


oxisto:
Meh, seems GitHub Actions has severe problems without go modules, So I will stick with having Go version with modules on Github Actions and the legacy versions on travis, which we will probably phase out at some point.

lggomez:
LGTM, although the minimum would be bumped to 1.13 per #9 in its current state

lggomez (approved):

PR #15, opened 2021-05-28
Backwards-compatible implementation of RFC7519's registered claim's structure

### LINKED ISSUE #11: Compliance with published RFC7519
When the library was initially started, [RFC7519](https://datatracker.ietf.org/doc/html/rfc7519) did not exist - only its drafts. This has lead to some shortcomings over the years where the implementation, especially of the claims, diverged from the now published RFC.

For example, the draft at the time of the writing of the claims struct actually stated that `iat`, `exp` and so on must be an `IntDate`, which did not include the possibility of non-integer date representations. It got then later changed to a `NumericDate` format in draft version 26, which - as @dunglas correctly pointed out - included also float representations (also strings).

The aim would be to be as backwards compatible as possible, e.g. with the introduction of  a new struct as suggested by @mfridman, which I will name for the lack of a better name `jwt.RFC7159Claims` for now.

This also was the original problem behind the audience mixup in #6

Points to consider (and possibly replaced with a new struct or function):

- [ ] The fields `ExpiresAt`, `IssuedAt`, `NotBefore` in `jwt.StandardClaims` are int64 instead of a (to be defined) `NumericDate` type. Probably solvable with a new struct
- [ ] The functions `VerifyIssuedAt`, `VerifyNotBefore`, `VerifyIssuedAt` of `jwt.StandardClaims` only compare against an int64 instead of the `NumericDate`. Also probably solvable with a new struct
- [ ] The functions `VerifyIssuedAt`, `VerifyNotBefore`, `VerifyIssuedAt` of `jwt.MapClaims` only compare against an int64 instead of the `NumericDate`. Trickier. I do not want to touch `jwt.MapClaims` too much. Probably the introduction of a new set of functions can be made and the old ones set to deprecated.
- [ ] `Audience` of `jwt.StandardClaims` should be a `[]string` instead of `string`. Could also be part of the new struct

Anything else that comes to mind?

At some point, it would also probably a good idea to check, whether [RFC8725](https://datatracker.ietf.org/doc/html/rfc8725) is also properly reflected in the implementation.

_Originally posted by @oxisto in https://github.com/golang-jwt/jwt/pull/9#discussion_r641039944_

oxisto:
This PR aims at implementing compliance to RFC7519, as documented in #11 without breaking the public API. It creates a new struct `RegisteredClaims` and deprecates (but not removes) the `StandardClaims`. It introduces a new type `NumericDate`, which represents a JSON numeric date value as specified in the RFC. This allows us to handle float as well as int-based time fields in `aud`, `exp` and `nbf`. Additionally, it introduces the type `StringArray`, which is basically a wrapper around `[]string` to deal with the oddities of the JWT `aud` field.

Closes #11


oxisto:
Now this introduces a whole range of other problems, specifically: Should we serialise the time as float including the fractions or not? There seems to a lot of confusion about that particular sentence in the RFC, whether actually any form of timestamp like `12345678.123456` should be supported or actually just things like `12345678.000000`

Not sure about the state of other JWT libraries, I fear that if we start to serialise as float, this might break more things in the end than it does good.

oxisto:
There is still a weird flaky failure on Go 1.15, but otherwise, this would be ready for a first review. @mfridman did you have something like this in mind?

lggomez:
> Now this introduces a whole range of other problems, specifically: Should we serialise the time as float including the fractions or not? There seems to a lot of confusion about that particular sentence in the RFC, whether actually any form of timestamp like `12345678.123456` should be supported or actually just things like `12345678.000000`

In this regard, RFC 3339 which is supersedes ISO 8601 as the standard for the HTTP date data type clearly states that dates must be serialized at complete precision, even at units of magnitude with zero value or in the absence of a timezone
 
> Not sure about the state of other JWT libraries, I fear that if we start to serialize as float, this might break more things in the end than it does good.

Well, that's the ugly side of standards. As developers we still cope with the misuse of IEEE 754 and its border cases that are _by design_ and using it in scenarios which are not clearly suitable for it like currencies. Enforcing standards can have unexpected side effects, so I get that in this scenario we do not want to break existing projects, and I agree to support extra precisions only if wanted. Kinda contradicts my previous point but it is a way to maintain the backwards compatibility in this current version

mfridman:
> Enforcing standards can have unexpected side effects, so I get that in this scenario we do not want to break existing projects, and I agree to support extra precisions only if wanted. Kinda contradicts my previous point but it is a way to maintain the backwards compatibility in this current version

In the current version, as long as we don't change `StandardClaims` and add a new struct it will be the consumer's choice which implementation they require.

Now whether to use an `int64`, `float64` or an embedded time for the new `RegisteredClaims` struct ?

Maybe an unpopular opinion, but I think the default should be the lowest allowed precision (12345678.000000) as this likely causes the least interoperability problems. 

To satisfy the requirement for fractions of a second precision, the concept of a `TimePrecision` variable is a nice middle ground.

> Not sure about the state of other JWT libraries, I fear that if we start to serialise as float, this might break more things in the end than it does good.

Curious, if the default is to ignore precision, unless explicitly set, why do you think it'd break things?

Couldn't we still handle float _and_ int-based time all the same?

oxisto:
> > Enforcing standards can have unexpected side effects, so I get that in this scenario we do not want to break existing projects, and I agree to support extra precisions only if wanted. Kinda contradicts my previous point but it is a way to maintain the backwards compatibility in this current version
> 
> In the current version, as long as we don't change `StandardClaims` and add a new struct it will be the consumer's choice which implementation they require.
> 
> Now whether to use an `int64`, `float64` or an embedded time for the new `RegisteredClaims` struct ?

I decided to go for an embedded time, since it seems the most "Go"-way to do it. It was actually also the way the original author tackled this problem in the `v4` branch, so I would say this is the most appropriate way

> Maybe an unpopular opinion, but I think the default should be the lowest allowed precision (12345678.000000) as this likely causes the least interoperability problems.

Yep, that it is now it is implemented now. `TImePrecision = time.Seconds`, which effectively generates integer timestamps, because 0 trailing digits of floats are cut in the Go formatting. 

> To satisfy the requirement for fractions of a second precision, the concept of a `TimePrecision` variable is a nice middle ground.
> 
> > Not sure about the state of other JWT libraries, I fear that if we start to serialise as float, this might break more things in the end than it does good.
> 
> Curious, if the default is to ignore precision, unless explicitly set, why do you think it'd break things?

It shouldn't now. I made the statement before introducing the time precision concept.

> Couldn't we still handle float _and_ int-based time all the same?

Basically, it is done this way now by having a `time.TIme` struct, which will parse either float or int as a Go date. All other comparisons are then done using the time struct.

mfridman:
> I decided to go for an embedded time, since it seems the most "Go"-way to do it. It was actually also the way the original author tackled this problem in the v4 branch, so I would say this is the most appropriate way

The time struct feels right, as most of the heavy lifting is done by the library instead of returning a basic type the caller would _anyways_ convert to a time object.

As you commented, I'm also curious if this would satisfy #9

oxisto:
@dunglas Did you have a chance to check, whether this would be a suitable replacement for #9?

james-d-elliott:
> Not sure about the state of other JWT libraries, I fear that if we start to serialise as float, this might break more things in the end than it does good.

Just as a personal opinion, there will be likely be unmarshal issues going forward if you do this. I would suggesting tabling this for v4 per semver rules or implementing the new structs etc. as suggested by @mfridman (i.e. *new* functionality instead of *breaking* functionality).

oxisto:
> > Not sure about the state of other JWT libraries, I fear that if we start to serialise as float, this might break more things in the end than it does good.
> 
> Just as a note though you're probably aware, there will be likely be unmarshal issues going forward if you do this. I would suggesting tabling this for v4 per semver rules or implementing the new structs etc. as suggested by @mfridman (i.e. _new_ functionality instead of _breaking_ functionality).

Just two things to bring to the discuss
1. The entire patch is based on *new* functionality (`RegisteredClaims`. It will not break any functionality, as long as you use the *old* struct (`StandardClaims`). So in this sense, it fully complies to the server rules
2. The default for the new functionality is `TimePrecision = time.Seconds`, meaning that no floats will be serialised, but pure integers as to not break too much even if you are using the *new* struct.

james-d-elliott:
Yep this makes sense and I was just giving some information regarding an experience we had previously with another library that changed from int to float64.

We will pickup any specific issues that affect us downstream anyway in integration testing most likely. So if we have issues when upgrading we can just hold off and figure out what to do.

mfridman:
Thanks for the feedback @james-d-elliott. I think the code will be easier to maintain and extend using the `NumericDate` (in this PR) as opposed to a `float64`

https://github.com/golang-jwt/jwt/pull/15/files#diff-1be21526f428f6177db963397455f724c264413a9abca372e95efa0a0318001aR21-R25

@oxisto Let's give this another week or two, and bearing no objections go with this approach in favor of https://github.com/golang-jwt/jwt/pull/9 ? 

I don't think we need two different implementations of an RFC7519-compatible `RegisteredClaims` struct.

oxisto:
> Thanks for the feedback @james-d-elliott. I think the code will be easier to maintain and extend using the `NumericDate` (in this PR) as opposed to a `float64`
> 
> https://github.com/golang-jwt/jwt/pull/15/files#diff-1be21526f428f6177db963397455f724c264413a9abca372e95efa0a0318001aR21-R25
> 
> @oxisto Let's give this another week or two, and bearing no objections go with this approach in favor of #9 ?
> 
> I don't think we need two different implementations of an RFC7519-compatible `RegisteredClaims` struct.

Yes, I think we can wait for another week or so, this is not completely time-critical. Sorry for spam-mentioning @dunglas here, but getting his opinion on this approach would be very helpful. I can understand if people are quite busy at the moment due to beginning holidays, etc. Otherwise, I think we can safely go ahead with this approach, we had quite a good discussion on the options here and choose a very "safe" approach in my opinion.

dunglas:
> @dunglas Did you have a chance to check, whether this would be a suitable replacement for #9?

I just reviewed your patch, and it looks good to me. I didn't review the details in depth, but I confirm that the general design is compatible with our needs.
Thank you for your work! I'll close #9.

oxisto:
> > @dunglas Did you have a chance to check, whether this would be a suitable replacement for #9?
> 
> I just reviewed your patch, and it looks good to me. I didn't review the details in depth, but I confirm that the general design is compatible with our needs.
> Thank you for your work! I'll close #9.

Excellent, thanks for having a look!

claims.go
```diff
@@ -12,9 +12,116 @@ type Claims interface {
...
+// This type can be used on its own, but then additional private and
+// public claims embedded in the JWT will not be parsed. The typical usecase
+// therefore is to embedded this in a user-defined claim type.
+//
+// See examples for how to use this with your own claim types.
+type RegisteredClaims struct {
+	// the `iss` (Issuer) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1
+	Issuer string `json:"iss,omitempty"`
+
+	// the `sub (Subject) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2
```
mfridman, line 28: nit. missing closing backtick .. `sub`

claims.go
```diff
@@ -12,9 +12,116 @@ type Claims interface {
...
+	Subject string `json:"sub,omitempty"`
+
+	// the `aud` (Audience) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3
+	Audience []string `json:"aud,omitempty"`
```
mfridman, line 32: Curious, have you considered this being a custom type that supported both `string` and `[]string`

> In the general case, the "aud" value is an array of case-
   sensitive strings, each containing a StringOrURI value.  In the
   special case when the JWT has one audience, the "aud" value MAY be a
   single case-sensitive string containing a StringOrURI value.

We could then implement custom unmarshalling to accommodate both. From the library perspective, it'd be an implementation detail, but to the caller we'd continue to expose `[]string`, for example:

json `string` or `[]string` -> go `[]string`

Or is `[]string` good enough for most use-cases, and if a user _must_ have it as a `string` they implement their own claims?

claims.go
```diff
@@ -12,9 +12,116 @@ type Claims interface {
...
+	if vErr.valid() {
+		return nil
+	}
+
+	return vErr
+}
+
+// VerifyAudience compares the aud claim against cmp.
+// If required is false, this method will return true if the value matches or is unset
+func (c *RegisteredClaims) VerifyAudience(cmp string, req bool) bool {
```
mfridman, line 82: If there are multiple audiences, would the caller just call this method in a loop? 

Or should there be a convenience method like:

`VerifyAudiences(aud []string, req bool)`

claims.go
oxisto: > Curious, have you considered this being a custom type that supported both `string` and `[]string`

I have seen this in other libraries and I did not really see the use case for it to be honest.

> 
> > In the general case, the "aud" value is an array of case-
> > sensitive strings, each containing a StringOrURI value.  In the
> > special case when the JWT has one audience, the "aud" value MAY be a
> > single case-sensitive string containing a StringOrURI value.
> 
> We could then implement custom unmarshalling to accommodate both. From the library perspective, it'd be an implementation detail, but to the caller we'd continue to expose `[]string`, for example:
> 
> json `string` or `[]string` -> go `[]string`
> 
> Or is `[]string` good enough for most use-cases, and if a user _must_ have it as a `string` they implement their own claims?

As far as I know, even if the aud is just a string instead of an array, it should still be correctly deserialised into a []string, so not sure if there is a use case of having just a string.

claims.go
oxisto: I would argue that while a JWT can be designed for multiple audiences, the check for an audience should check whether the callee's audience, e.g. its service URI, is within that list. Not sure if it makes sense to have multiple audiences in a single service.

claims.go
mfridman: > As far as I know, even if the aud is just a string instead of an array, it should still be correctly deserialised into a []string, so not sure if there is a use case of having just a string.

If the incoming JWT "aud" claim is a string (and not an array of strings) deserialization will fail:

```
json: cannot unmarshal string into Go struct field RegisteredClaims.aud of type []string
```

But this is fine, disregard this as the goal of this `RegisteredClaims` is to support the most common use case.

I wish the RFC just omitted this "special case" altogether.

oxisto: I have added a custom type now, named `StringArray`. Not sure about the naming, anyone got any good ideas?

oxisto:
Looks like I broke GitHub. All repository fetches return 500. I will try to re-run the CI later.

dcormier:
Closes #32.

claims.go
```diff
@@ -12,9 +12,116 @@ type Claims interface {
...
+	Issuer string `json:"iss,omitempty"`
+
+	// the `sub` (Subject) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2
+	Subject string `json:"sub,omitempty"`
+
+	// the `aud` (Audience) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3
+	Audience StringArray `json:"aud,omitempty"`
+
+	// the `exp` (Expiration Time) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4
+	ExpiresAt *NumericDate `json:"exp,omitempty"`
```
amnonbc, line 35: why have pointers to a NumericDate, rather than a NumericDate? Why do we need the `*`?

types.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// UnmarshalJSON is an implementation of the json.RawMessage interface and deserializses a
+// NumericDate from a JSON representation, i.e. a json.Number. This number represents an UNIX epoch
+// with either integer or non-integer seconds.
+func (date *NumericDate) UnmarshalJSON(b []byte) (err error) {
+	var (
+		number json.Number
+		f      float64
+	)
```
amnonbc, line 55:
```suggestion
	var	number json.Number
	var	f      float64
```
Grouping the two vars together looks weird.

types.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err = json.Unmarshal(b, &number); err != nil {
+		return fmt.Errorf("could not parse NumericData: %w", err)
+	}
```
amnonbc, line 59:
```suggestion
	err := json.Unmarshal(b, &number)
	if err != nil {
		return fmt.Errorf("could not parse NumericData: %w", err)
	}
```
is easier to read...

amnonbc (commented):
This PR would be simpler if we changed `RegisteredClaims.ExpiresAt` `NotBefore` and `IssuedAt` to `NumericDate` rather than `*NumericDate`.
If the fields are not defied. these fields would then have the zero value.

oxisto:
> This PR would be simpler if we changed `RegisteredClaims.ExpiresAt` `NotBefore` and `IssuedAt` to `NumericDate` rather than `*NumericDate`.
> If the fields are not defied. these fields would then have the zero value.

This probably comes down to individual gusto, but I would argue that it is easier to see whether the fields are set or not if a pointer is used and set to `nil`, rather than setting the variable to something that resembles "zero". Is there any particular downside to using a pointer here from the user perspective?

claims.go
mfridman: Apologies for the delay.

Can you please add a net new test for parsing `RegisteredClaims`, specifically a test that uses a new JWT with multiple audience claims? Afaics none of the tests have JWTs with multiple audiences. In the `UnmarshalJSON` method might need to switch on `[]interface{}` and assert for `string`.. tests should help here.

Also, what will happen if a user of this package needs to _generate_ a token with "aud" as a `string` instead of `[]string`? I.e., they need to return `{ "aud": "example.com" }`. Is it okay to point users at creating their own custom claims, or should the library natively support (de)-serializing "aud" claim both ways?

oxisto: Sorry for not progressing on this, I was quite busy at work, I hopefully have some time during the coming weekend to finalise this PR.

mfridman: Of course, it's open source and we try to do what we can :) 

The library is in a stable state and much of what we're adding is incremental improvements, so it's natural it'll take time.

oxisto:
@mfridman, I have added the suggestions you made, the tests, plus I made the serialisation configurable. I am still not quite happy with the name `StringArray`, sounds more like Java than Go, so I am open for suggestions here.

claims.go
oxisto: Probably comes down to personal preferences a little bit, although with (#40) in mind, I like the idea that we can actually differentiate between scenarios where a value is not set or contains unexpected types (thus being `nil` + an error) vs. the case where (for whatever reason) the date has the value `0`.

oxisto:
> @mfridman, I have added the suggestions you made, the tests, plus I made the serialisation configurable. I am still not quite happy with the name `StringArray`, sounds more like Java than Go, so I am open for suggestions here.

I have renamed it to `ClaimStrings`, this was also the name used in the original `v4` and I quite like it. I think all the remaining issues have now been solved.

dunglas:
I tried to use the latest version of this patch in Mercure, and it works like a charm: https://github.com/dunglas/mercure/pull/545

+1 on my side to merge.

oxisto:
I merged in v4, there were some conflicts but everything should be ready now.

oxisto:
I know everyone is busy (including myself). Giving a small nudge to @mfridman, @lggomez, @Waterdrips, @ripienaar. There are several request for this coming up in the issues. However it is a big PR so I definitely want to have some reviews here before we merge it :)

lggomez (changes requested):
I'd look to replicate #83 into this new implementation, otherwise it LGTM

mfridman:
> I know everyone is busy (including myself). Giving a small nudge to @mfridman, @lggomez, @Waterdrips, @ripienaar. There are several request for this coming up in the issues. However it is a big PR so I definitely want to have some reviews here before we merge it :)

Apologies, I have been quite busy as of late. I'll have some more time this week and can take a look if others haven't.

types.go
```diff
@@ -0,0 +1,127 @@
...
+
+// TimePrecision sets the precision of times and dates within this library.
+// This has an influence on the precision of times when comparing expiry or
+// other related time fields. Furthermore, it is also the precision of times
+// when serializing.
+//
+// For backwards compatibility the default precision is set to seconds, so that
+// no fractional timestamps are generated.
+//
+// TODO(oxisto): the tests seem to fail sometimes, if the precision is microseconds because the difference is literally 1 microsecond
```
lggomez, line 19: Is there anything further you could pinpoint on this? from seeing the implementation I'd put my money on a IEEE rounding error but that should be reproducible and since it is a seemingly random microsecond error on the same test it likely is a clock skew

time.Unix doesn't involve any float aritmetic https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/time/time.go;l=1317-1328

the likely cause of the skew is time.Truncate, which does indeed some extensive calculations on the division and then strips the monotonic clock reading (time.ext) https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/time/time.go;l=1442-1453;drc=refs%2Ftags%2Fgo1.17. Additionally, time.UnixNano() internally applies additional truncation and uses both wall and ext time

types.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return
+}
+
+func (s ClaimStrings) MarshalJSON() (b []byte, err error) {
+	// This handles a special case in the JWT RFC. If the string array, e.g. used by the "aud" field,
+	// only contains one element, it MAY be serialized as a single string. This may or may not be
+	// desired based on the ecosystem of other JWT library used, so we make it configurable by the
+	// variable MarshalSingleStringAsArray.
+	if len(s) == 1 && !MarshalSingleStringAsArray {
```
mfridman, line 122: great, I like this behaviour.

types.go
mfridman: I think it's safe to remove this TODO and comment.

If users are setting precision of 1 μs this isn't even applicable in real-world systems, esp. given how hard it is to coordinate time in distributed systems. 

Tbh I'd even bump the tests to something like 1ms or even 10ms.

mfridman (approved):
lgtm.

oxisto:
> I'd look to replicate #83 into this new implementation, otherwise it LGTM

Done

lggomez (dismissed):
LGTM

oxisto:
Should be good to go then, I had to resolve one last merge conflict. Sorry @lggomez this dismissed your review for some reason... I need another approve for it to merge :)

mfridman:
Lgtm

mfridman (approved):

mfridman:
@oxisto I saw a few mentions to this ticket, are we safe to tag a new release `4.1.0` ? 

This was on of the main changes since the prior release, figured I'd ask just in case.

oxisto:
> @oxisto I saw a few mentions to this ticket, are we safe to tag a new release `4.1.0` ?
> 
> This was on of the main changes since the prior release, figured I'd ask just in case.

We might need to wait until we resolve #86

mfridman:
I'm not sure how best to coordinate releases, maybe discussions or a tracking "releases" issue?

Just confirming now that #86 is landed we can cut a release: `4.1.0`

Issue #17, opened 2021-05-29
Address cached references to old versions in the Go module proxy


mfridman:
Unfortunately the Go module proxy (i.e., mirror) cached 2 versions:

https://pkg.go.dev/github.com/golang-jwt/jwt?tab=versions

When a user does a `go get github.com/golang-jwt/jwt` it will download the following version:

    github.com/golang-jwt/jwt v3.2.0+incompatible // indirect

Whereas it _should_ import the correct one (based on latest commit on `main`)

    github.com/golang-jwt/jwt v0.0.0-20210529012641-6a07921e6808 // indirect

If we cannot get those 2 versions removed from the proxy, we'll have to go with plan B... add a /v3 and tag a `v3.2.1`

oxisto:
Did you already file an issue for that?

oxisto:
> What _could_ also work is to retract the two versions: https://blog.golang.org/go116-module-changes#TOC_5. But I think this probably Is only going to work on Go 1.16.
> 
> Not a good start :(

As can be seen in https://github.com/golang-jwt/jwt/pull/23 module retraction is in no way backwards compatible...

oxisto:
Sorry for the fragmented thoughts and posts here... With all the information available currently, I would propose the following action plan

## Step 1: Remove `go.mod` again

Currently, we have no external dependencies in the library (this is also probably a good reason why this was so popular). So the benefit of having go modules is debatable and almost zero. With the exception of the annoying `-compatible` branding on the version - which bugs me as well - but I think we could live with that for a moment. Go's way of enforcing SIV after a `v2` has lead to so much debate and pain, but unfortunately the design seems frozen these days.

## Step 2: Release a v3.2.1+incompatible

Mainly with #6 fixed, possibly also #15. 
Optional: Re-release the old release tags, so that we have the complete version history

## Step 3: Continue working on backwards-compatible stuff in the `v3.x.x+incompatible` branch

## Step 4: Prepare a `v4` branch with Go modules active

This branch will introduce breaking changing to the API as well as the import path, as we will switch to `github.com/golang-jwt/jwt/v4` and do a proper release with Go modules.

mfridman:
I created an issue, but it'll likely be closed. https://github.com/golang/go/issues/46440

Since users would need to modify their import paths _anyways_ couldn't we do the following:

1. Re-release the old release tags, so that we have the complete version history
2. Modify go.mod and suffix the module with /v3, adding SIV 
[...]

PR #18, opened 2021-05-29
WIP: Add golanci/golint to github actions

Waterdrips:
Add golangci linting to ensure consistent code standards.

Signed-off-by: Alistair Hey <alistair@heyal.co.uk>

### DIFF
--- a/http_example_test.go
+++ b/http_example_test.go
@@ -32,9 +32,9 @@ var (
 	serverPort int
 	// storing sample username/password pairs
 	// don't do this on a real server
-	users = map[string]string{
-		"test": "known",
-	}
+	//users = map[string]string{
+	//	"test": "known",
+	//}
 )
 
 // read the key files before starting http handlers
@@ -65,7 +65,7 @@ func init() {
 	}()
 }
 
-var start func()
+//var start func()
 
 func fatal(err error) {
 	if err != nil {
@@ -101,8 +101,12 @@ func Example_getTokenViaHTTP() {
 
 	// Read the token out of the response body
 	buf := new(bytes.Buffer)
-	io.Copy(buf, res.Body)
-	res.Body.Close()
+	if _, err = io.Copy(buf, res.Body); err != nil {
+		fatal(err)
+	}
+	if err = res.Body.Close(); err != nil {
+		fatal(err)
+	}
 	tokenString := strings.TrimSpace(buf.String())
 
 	// Parse the token
@@ -136,8 +140,12 @@ func Example_useTokenViaHTTP() {
 
 	// Read the response body
 	buf := new(bytes.Buffer)
-	io.Copy(buf, res.Body)
-	res.Body.Close()
+	if _, err = io.Copy(buf, res.Body); err != nil {
+		panic(err)
+	}
+	if err = res.Body.Close(); err != nil {
+		panic(err)
+	}
 	fmt.Println(buf.String())
 
 	// Output: Welcome, foo
@@ -214,5 +222,4 @@ func restrictedHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Token is valid
 	fmt.Fprintln(w, "Welcome,", token.Claims.(*CustomClaimsExample).Name)
-	return
 }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,3 +26,12 @@ jobs:
           go vet ./...
           go test -v ./...
           go build ./...
+  lint:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.29
--- a/ecdsa.go
+++ b/ecdsa.go
@@ -13,7 +13,7 @@ var (
 	ErrECDSAVerification = errors.New("crypto/ecdsa: verification error")
 )
 
-// Implements the ECDSA family of signing methods signing methods
+// SigningMethodECDSA Implements the ECDSA family of signing methods signing methods
 // Expects *ecdsa.PrivateKey for signing and *ecdsa.PublicKey for verification
 type SigningMethodECDSA struct {
 	Name      string
@@ -53,7 +53,7 @@ func (m *SigningMethodECDSA) Alg() string {
 	return m.Name
 }
 
-// Implements the Verify method from SigningMethod
+// Verify Implements the Verify method from SigningMethod
 // For this verify method, key must be an ecdsa.PublicKey struct
 func (m *SigningMethodECDSA) Verify(signingString, signature string, key interface{}) error {
 	var err error
@@ -85,7 +85,9 @@ func (m *SigningMethodECDSA) Verify(signingString, signature string, key interfa
 		return ErrHashUnavailable
 	}
 	hasher := m.Hash.New()
-	hasher.Write([]byte(signingString))
+	if _, err = hasher.Write([]byte(signingString)); err != nil {
+		return err
+	}
 
 	// Verify the signature
 	if verifystatus := ecdsa.Verify(ecdsaKey, hasher.Sum(nil), r, s); verifystatus {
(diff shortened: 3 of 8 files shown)


Waterdrips:
adding error checking may have uncovered a bug - going to take a look a little later

http_example_test.go
```diff
@@ -101,8 +101,12 @@ func Example_getTokenViaHTTP() {
 
 	// Read the token out of the response body
 	buf := new(bytes.Buffer)
-	io.Copy(buf, res.Body)
-	res.Body.Close()
+	if _, err = io.Copy(buf, res.Body); err != nil {
+		fatal(err)
+	}
+	if err = res.Body.Close(); err != nil {
```
amnonbc, line 107: Not very useful to panic in this situation.
An error when closing a response.Body can be ignored - the client has gone anyway.

This is an example where using golangci-lint does not improve the code, 
but makes it unnecessarily convoluted.

The Go stdlib wil notpass golangci-lint.

amnonbc (commented):
This PR does not improve the code.
The choices made by golangci are questionable.

I would run go vet, but not golangci.

oxisto:
> This PR does not improve the code.
> The choices made by golangci are questionable.
> 
> I would run go vet, but not golangci.

I agree with you, that golangci should be taken with a grain of salt. So maybe it would make sense to start with a `go vet` approach, fix the most "critical" code smells and go from there. Would you mind preparing such a PR with just `go vet` for now? We can then see how much the 'delta' is going from go vet to golangci.

> 
> Also I would remove support for any versions < 1.15

This is something that we need to be very careful about in my opinion. The original project did not have a `go.mod` file and we also do not have one (for now, see #17 for a longer discussion on that), until we go for a `v4` branch. So it is quite hard for us to understand what the minimum required go version was for most people using it. We want to provide a smooth transition path and will probably continuously phase out older Go versions as we introduce new PRs that rely on new language features / constructs.

amnonbc:
> > This PR does not improve the code.
> > The choices made by golangci are questionable.
> > I would run go vet, but not golangci.
> 
> I agree with you, that golangci should be taken with a grain of salt. So maybe it would make sense to start with a `go vet` approach, fix the most "critical" code smells and go from there. Would you mind preparing such a PR with just `go vet` for now? We can then see how much the 'delta' is going from go vet to golangci.

We should fix all the warning from go vet.

And I would argue that golangci should be abandoned, as it includes too many questionable
tests, and has a very low signal to noise ratio.


> 
> > Also I would remove support for any versions < 1.15
> 
> This is something that we need to be very careful about in my opinion. The original project did not have a `go.mod` file and we also do not have one (for now, see #17 for a longer discussion on that), until we go for a `v4` branch. So it is quite hard for us to understand what the minimum required go version was for most people using it. We want to provide a smooth transition path and will probably continuously phase out older Go versions as we introduce new PRs that rely on new language features / constructs.

The current Go version is 1.16. 1.17 is now in public beta and should be released within a month or two.
1.15 is still maintained. And anything below 1.14 is no longer supported (even for security fixes).
Nobody should be using a version of Go below 1.15.
So I do not understand what you mean by "minimum version requirement". Anybody using an obsolete version 
of Go should upgrade. There is no excuse not to.
And if anyone is not prepared to upgrade the Go toolchain, then why would they want to upgrade their jwt library?
And why should we support such irrationality?

oxisto:
> > > This PR does not improve the code.
> > > The choices made by golangci are questionable.
> > > I would run go vet, but not golangci.
> > 
> > 
> > I agree with you, that golangci should be taken with a grain of salt. So maybe it would make sense to start with a `go vet` approach, fix the most "critical" code smells and go from there. Would you mind preparing such a PR with just `go vet` for now? We can then see how much the 'delta' is going from go vet to golangci.
> 
> We should fix all the warning from go vet.
> 
> And I would argue that golangci should be abandoned, as it includes too many questionable
> tests, and has a very low signal to noise ratio.
> 
> > > Also I would remove support for any versions < 1.15
> > 
> > 
> > This is something that we need to be very careful about in my opinion. The original project did not have a `go.mod` file and we also do not have one (for now, see #17 for a longer discussion on that), until we go for a `v4` branch. So it is quite hard for us to understand what the minimum required go version was for most people using it. We want to provide a smooth transition path and will probably continuously phase out older Go versions as we introduce new PRs that rely on new language features / constructs.
> 
> The current Go version is 1.16. 1.17 is now in public beta and should be released within a month or two.
> 1.15 is still maintained. And anything below 1.14 is no [longer supported](https://golang.org/doc/devel/release#policy) (even for security fixes).
> Nobody should be using a version of Go below 1.15.
> So I do not understand what you mean by "minimum version requirement". Anybody using an obsolete version
> of Go should upgrade. There is no excuse not to.
> And if anyone is not prepared to upgrade the Go toolchain, then why would they want to upgrade their jwt library?
> And why should we support such irrationality?

I am not trying to start a heated discussion about versioning policies here, but explain a little bit the reasoning behind my remarks.

Me, *personally*: I am a big favour of always updating as soon as possible possible and I am a big fan of dependabot, renovate, etc. to always pull in the newest versions (decent test coverage needed of course).

Me, as one of the *maintainers* of this library: I try to not be opinionated about the usage of this library or the constraints the users might have with regards to versions. The goal was (and still is) to provide a 1:1 drop-in replacement for https://github.com/dgrijalva/jwt-go while fixing some of the known issues. Since we are taking "over" existing work, we wanted to follow the ideas of the original author (and other contributors) as close as possible. Without reading their minds, that is a hard thing to do, so a lot of it is some guess work or trying to read what was written in the various documentation files. We have done that with the `v3.2.1` release.

What I meant with "minimum version requirement" is that the `jwt-go` library never specifically mentioned any minimum Go version, especially because no `go.mod` file is present. From the point of view of the code, the minimum version that still compiles is 1.7. I agree with you, that it is not a "good" thing to do, to run this old Go version. But once again, I am trying to not be opinionated about it. 

Now, as said before, moving forward, we will most likely drop older versions, mainly because we want to use features of the newer languages.

I hope that clarifies my standpoint here a little bit better.

amnonbc:
1) Go is an opinionated language, and the Go community is opinionated when it comes to supported versions.
There is zero benefit in using obsolete versions or supporting them.
More concretely, continuing to support 1.14 and earlier prevents us merging 
https://github.com/dgrijalva/jwt-go/pull/455 which fixes the library's pathologically slow handling of 
ECDSA verification.

ripienaar:
It's common to support only the 2 most recent major releases of Go.

I can't pretend to have studied all of [these vulnerabilities](https://www.cvedetails.com/vulnerability-list/vendor_id-14185/Golang.html) - but Go do have a fair few vulnerabilities in the langauge and I'd rather signal our support of newer versions via go mod.

We made a release that fixes the vulnerability and provided a drop in replacement, seems reasonable to tighten things up moving forward to me

ripienaar:
Regarding linters, go vet is decent others is noisy and questionable, used to like staticcheck but even that is turning into a rubocop like cancer causing constant churn in code bases for no good reason.

amnonbc:
> It's common to support only the 2 most recent major releases of Go.
> 
> I can't pretend to have studied all of [these vulnerabilities](https://www.cvedetails.com/vulnerability-list/vendor_id-14185/Golang.html) - but Go do have a fair few vulnerabilities in the langauge and I'd rather signal our support of newer versions via go mod.
> 
> We made a release that fixes the vulnerability and provided a drop in replacement, seems reasonable to tighten things up moving forward to me

Yes, we have to take security very seriously in a JWE library.
We can not base the implementation on Go versions that no longer get security fixes.

lggomez:
> Regarding linters, go vet is decent others is noisy and questionable, used to like staticcheck but even that is turning into a rubocop like cancer causing constant churn in code bases for no good reason.

I agree that linters an be a PITA but I like to see the static analysis side of it as many of their analyzers aren't style only and help catch bugs before even showing at runtime. How bureaucratic the process is is up to the linter's configurability and the codebase's configuration

That said, I'd like to continue with some static analysis step (be it golangci-lint or staticcheck) but with a custom configuration for this project that suits the codebase and marks the really important findings

mfridman:
Historically I've used/encountered `staticcheck` in most projects, my vote goes to `staticcheck`. But I don't have a strong opinion... `golangci-lint` is just fine too.

One way or another, having _something_ is probably better than nothing at least to conform to some "style", which we'll probably figure out with time.

oxisto:
> Historically I've used/encountered `staticcheck` in most projects, my vote goes to `staticcheck`. But I don't have a strong opinion... `golangci-lint` is just fine too.
> 
> One way or another, having _something_ is probably better than nothing at least to conform to some "style", which we'll probably figure out with time.

I played around with `staticcheck` in another PR (#44). Looks ok so far, but it seems that not even the mismatched documentation strings are caught with just this approach; which is weird because I thought that `go vet` would catch it.

Waterdrips:
closing because it was contentious.

PR #23, opened 2021-05-29
Added module retraction for version v1.0.2 and v3.2.0-incompatible

oxisto:
It seems that two old versions were indexed by the golang mirror. This
will retract them, but I fear that this only works for Go 1.16.


oxisto:
Hmpf, it seems, this is not suitable at all, because on older Go versions, this won't even compile because of the keyword.

PR #27, opened 2021-06-04
Update VERSION_HISTORY.md and MIGRATION_GUIDE.md

### LINKED ISSUE #25: update MIGRATION_GUIDE and VERSION_HISTORY
https://github.com/golang-jwt/jwt/blob/master/MIGRATION_GUIDE.md 

Is currently talking about migrating from v2 ro v3, this doesnt make much sense in this reop's context.

We could update this doc to discuss how to migrate from `dgrijalva/jwt-go/` to `golang-jwt/jwt` as a drop-in replacment, and then we can add sections in the future for new versions published from here

oxisto:
Fixes #25


oxisto:
@mfridman I think you had a suggesting for replacements before, would you mind adding that here? something like 

```
go mod edit -replace github.com/dgrijalva/jwt-go=github.com/golang-jwt/jwt@v3.2.1+incompatible
```

But will that work with our new import path? Or should we provide some `sed`-style replacement command?

MIGRATION_GUIDE.md
```diff
@@ -1,97 +1,22 @@
...
-		fmt.Printf("Token for user %v expires %v", claims.User, claims.StandardClaims.ExpiresAt)
-	}
+go mod edit -replace github.com/dgrijalva/jwt-go=github.com/golang-jwt/jwt@v3.2.1+incompatible
+go mod tidy
 ```
 
-### `ParseFromRequest` has been moved
-
-To keep this library focused on the tokens without becoming overburdened with complex request processing logic, `ParseFromRequest` and its new companion `ParseFromRequestWithClaims` have been moved to a subpackage, `request`.  The method signatues have also been augmented to receive a new argument: `Extractor`.
+This will still keep the old import path in your code but replace it with the new package and also introduce a new indirect dependency to `github.com/golang-jwt/jwt`. Try to compile your project; it should still work.
```
mfridman, line 14: Not sure if the section above or here is more appropriate, but we should mentione _somewhere_

The `v3.x.x+incompatible` lineage will continue to be backwards compatible. As a consumer of this package I'd appreciate a note of this.

mfridman (approved):
lgtm. I see the backwards-compatibility was mentioned in the README so that's good enough.

PR #28, opened 2021-06-20
remove support for Go <= 1.14

amnonbc:
Go 1.14 and below are no longer supported by the toolchain and standard libraries and no longer provide security updates.

See https://github.com/golang-jwt/jwt/issues/29


.travis.yml
```diff
@@ -5,7 +5,5 @@ script:
     - go test -v ./...
```
oxisto, line 5: The idea was to keep the Travis file for the older versions, removing them, once we (gradually) arrive at the newer versions, so in this case you could just delete it. We are using GitHub Actions as the primary CI. GH actions had issues with older Go versions, that is why the travis file still existed.

oxisto:
If we decide to do this, I would suggest having a clear notice about this in the README, since we are currently not able to go the `go.mod` route. Can you include such a sentence or notice?

.travis.yml
amnonbc: The fact that GH actions has issues with ancient Go versions, gives us a very strong hint that nobody is using them.

README.md
```diff
@@ -9,10 +9,19 @@ A [go](http://www.golang.org) (or 'golang' for search engine friendliness) imple
...
 
-**SECURITY NOTICE:** Some older versions of Go have a security issue in the crypto/elliptic. Recommendation is to upgrade to at least 1.8.3. See issue [dgrijalva/jwt-go#216](https://github.com/dgrijalva/jwt-go/issues/216) for more detail.
+**SECURITY NOTICE:** Some older versions of Go have a security issue in the crypto/elliptic. Recommendation is to upgrade to at least 1.15 See issue [dgrijalva/jwt-go#216](https://github.com/dgrijalva/jwt-go/issues/216) for more detail.
 
 **SECURITY NOTICE:** It's important that you [validate the `alg` presented is what you expect](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/). This library attempts to make it easy to do the right thing by requiring key types match the expected alg, but you should take the extra step to verify it in your usage.  See the examples provided.
 
+### Supported Go versions
+
+Our support of Go versions is alaigned with Go's [version release policy](https://golang.org/doc/devel/release#policy).
```
oxisto, line 18: minor spelling: alaigned

README.md
```diff
@@ -9,10 +9,19 @@ A [go](http://www.golang.org) (or 'golang' for search engine friendliness) imple
...
+So we will support a major version of Go unil there are two newer major releases.
```
oxisto, line 19: minor spelling: unil

README.md
```diff
@@ -9,10 +9,19 @@ A [go](http://www.golang.org) (or 'golang' for search engine friendliness) imple
...
+We no longer support building jwt-go with unsupported Go versions, as these contain secruity vulnerabilities
+which will not be fixed.
+This means that currently we support Go 1.15 and 1.16. Anybody using older versions of the toolchain should upgrade
```
oxisto, line 22: I would suggest removing "Anybody using older versions of the toolchain should upgrade to the latest stable release". I think it is already enough that we state we only support 1.15, 1.16. Let's not overdo it.

.travis.yml
oxisto: You can remove the `.travis.yml` then

oxisto (changes requested):
Only minor things like spelling need to be change please, Could you also maybe drop a hint in the `VERSION_HISTORY.md` about the v3.2.1 being the last version that supports Go <= 1.14? Otherwise, I will do that in a follow-up commit.

Apart from that, let's see what others think @mfridman, @Waterdrips 

As far as I could gather from the other conversation, @ripienaar was already in favour of this approach.

README.md
oxisto: Minor spelling: secruity

oxisto (dismissed):
Looks good so far, but I would really like the opinion of the rest of the team here, since this is a potentially "sensitive" issue.

README.md
```diff
@@ -9,10 +9,18 @@ A [go](http://www.golang.org) (or 'golang' for search engine friendliness) imple
...
 
 **SECURITY NOTICE:** It's important that you [validate the `alg` presented is what you expect](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/). This library attempts to make it easy to do the right thing by requiring key types match the expected alg, but you should take the extra step to verify it in your usage.  See the examples provided.
 
+### Supported Go versions
+
+Our support of Go versions is aligned with Go's [version release policy](https://golang.org/doc/devel/release#policy).
+So we will support a major version of Go until there are two newer major releases.
+We no longer support building jwt-go with unsupported Go versions, as these contain security vulnerabilities
+which will not be fixed.
+This means that currently we support Go 1.15 and 1.16.
```
ripienaar, line 22: We might drop the specific statement of 1.15 and 1.16 here else we end up having to always remember to update this - which we probably wont :)

Otherwise LGTM

ripienaar (dismissed):

mfridman (dismissed):
Agree with this [comment](https://github.com/golang-jwt/jwt/pull/28/files#r655924078), just remove references to specific versions.

Otherwise, I don't see an issue dropping support for older versions. (probably for the best anyways)

ripienaar (approved):

mfridman (approved):
lgtm. Thank you for your contribution.

ashish-scalent:
trying to retrieve context tokens from req after checkJWT middleware 
```
ctxToken := r.Context().Value(jwtmiddleware.ContextKey{})
```
using the above I am able to retrieve the tokens but that is an interface which is with type string 
previously it was *jwt.Tokens 

also previously we were fetching that using the 
ctxToken := r.Context().Value("user")

but now the user is filed is not there

PR #34, opened 2021-07-12
remove unnecessary for loop in token signing string for readability

hyeonjae:
remove unnecessary for-loop in token signing string for readability
and ...
 - add testcase
 - add benchmark
 - improve performance slightly

### DIFF
--- a/token_test.go
+++ b/token_test.go
@@ -70,6 +70,8 @@ func BenchmarkToken_SigningString(b *testing.B) {
 		Claims:    jwt.StandardClaims{},
 	}
 	b.Run("BenchmarkToken_SigningString", func(b *testing.B) {
+		b.ResetTimer()
+		b.ReportAllocs()
 		for i := 0; i<b.N; i++ {
 			t.SigningString()
 		}
--- a/token_test.go
+++ b/token_test.go
@@ -1,0 +1,77 @@
+package jwt_test
+
+import (
+	"testing"
+
+	"github.com/golang-jwt/jwt"
+)
+
+func TestToken_SigningString(t1 *testing.T) {
+	type fields struct {
+		Raw       string
+		Method    jwt.SigningMethod
+		Header    map[string]interface{}
+		Claims    jwt.Claims
+		Signature string
+		Valid     bool
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "",
+			fields: fields{
+				Raw:    "",
+				Method: jwt.SigningMethodHS256,
+				Header: map[string]interface{}{
+					"typ": "JWT",
+					"alg": jwt.SigningMethodHS256.Alg(),
+				},
+				Claims:    jwt.StandardClaims{},
+				Signature: "",
+				Valid:     false,
+			},
+			want: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30",
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t1.Run(tt.name, func(t1 *testing.T) {
+			t := &jwt.Token{
+				Raw:       tt.fields.Raw,
+				Method:    tt.fields.Method,
+				Header:    tt.fields.Header,
+				Claims:    tt.fields.Claims,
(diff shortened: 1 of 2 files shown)


oxisto (commented):
Thanks. Indeed improves readability a little bit. I wonder if there was any trick to the initial implementation that I am not seeing at first glance here.

token_test.go
```diff
@@ -0,0 +1,77 @@
...
+				return
+			}
+			if got != tt.want {
+				t1.Errorf("SigningString() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func BenchmarkToken_SigningString(b *testing.B) {
```
oxisto, line 63: Did you do a comparison between the old and the new approach? It may be the case, that this new approach slightly increases number of allocations, since you add addtional two variables for `claims` and `header` whereas the original implementation directly stores it into an array (which you need to create temporarily as well in the last call)

token_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t := &jwt.Token{
+		Method:    jwt.SigningMethodHS256,
+		Header: map[string]interface{}{
+			"typ": "JWT",
+			"alg": jwt.SigningMethodHS256.Alg(),
+		},
+		Claims:    jwt.StandardClaims{},
+	}
+	b.Run("BenchmarkToken_SigningString", func(b *testing.B) {
```
lggomez, line 72:
```suggestion
	b.Run("BenchmarkToken_SigningString", func(b *testing.B) {
		b.ResetTimer()
		b.ReportAllocs()
```

token_test.go
lggomez: I took the liberty to test both implementations and there doesn't seem to be a perf regression:

current

> goos: darwin
> goarch: amd64
> pkg: github.com/golang-jwt/jwt/v4
> cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
> BenchmarkToken_SigningString
> BenchmarkToken_SigningString/BenchmarkToken_SigningString
> BenchmarkToken_SigningString/BenchmarkToken_SigningString-12         	  820801	      1417 ns/op	     600 B/op	      17 allocs/op
> PASS

new

> goos: darwin
> goarch: amd64
> pkg: github.com/golang-jwt/jwt/v4
> cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
> BenchmarkToken_SigningString
> BenchmarkToken_SigningString/BenchmarkToken_SigningString
> BenchmarkToken_SigningString/BenchmarkToken_SigningString-12         	  771712	      1400 ns/op	     600 B/op	      17 allocs/op
> PASS

token_test.go
```diff
@@ -0,0 +1,77 @@
+package jwt_test
+
+import (
+	"testing"
+
+	"github.com/golang-jwt/jwt"
```
lggomez, line 6: Please rebase to main and updater the import path to `github.com/golang-jwt/jwt/v4` for this test file to work

lggomez (changes requested):
Looks good, needs an update from main and the requested change to make the benchmark to be more precise and report the allocations, otherwise I think we're fine (also see https://github.com/golang-jwt/jwt/pull/34/files#r680371882)

token_test.go
```diff
@@ -0,0 +1,79 @@
...
```
lggomez, line 6:
```suggestion
	"github.com/golang-jwt/jwt/v4"
```

lggomez (approved):

lggomez:
This is another stale code refactor that we'd want to have in v4 and before other refactors like #115 

LGTM. Care to give it a check @oxisto? since the PR author didn't respond these last months I commited the purposed changes to make this benchmarkable on v4

mfridman (approved):

oxisto:
> This is another stale code refactor that we'd want to have in v4 and before other refactors like #115
> 
> LGTM. Care to give it a check @oxisto? since the PR author didn't respond these last months I commited the purposed changes to make this benchmarkable on v4

I was still cautions to touch functions that are the very core of this library, but it looks sound, doesn't add more allocations and is marginally faster, so I think you can merge it.

oxisto (approved):

PR #36, opened 2021-07-27
add ed25519 support

vtolstov:
Signed-off-by: Vasiliy Tolstov <v.tolstov@unistack.org>


oxisto:
Thank you for porting this to the new repo!

ed25519_utils.go
```diff
@@ -0,0 +1,53 @@
+package jwt
+
+import (
+	"crypto"
+	"encoding/asn1"
+	"encoding/pem"
+
+	"golang.org/x/crypto/ed25519"
```
oxisto, line 8: Same as above

oxisto (changes requested):

ed25519_utils.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Parse PEM-encoded Edwards curve private key
+func ParseEdPrivateKeyFromPEM(key []byte) (crypto.PrivateKey, error) {
+	var block *pem.Block
+	if block, _ = pem.Decode(key); block == nil {
+		return nil, ErrKeyMustBePEMEncoded
+	}
+	var asn1PrivKey ed25519PrivKey
+	if _, err := asn1.Unmarshal(block.Bytes, &asn1PrivKey); err != nil {
```
oxisto, line 33: You could use something along these lines instead:

```
// Parse the key
var parsedKey interface{}
if parsedKey, err = x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
	return nil, err
}

var pkey ed25519.PrivateKey
var ok bool
if pkey, ok = parsedKey.(ed25519.PrivateKey); !ok {
	return nil, ErrNotEdPrivateKey
}

return pkey
```

This would avoid introducing new temporary structs.

ed25519_utils.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Parse PEM-encoded Edwards curve public key
+func ParseEdPublicKeyFromPEM(key []byte) (crypto.PublicKey, error) {
+	var block *pem.Block
+	if block, _ = pem.Decode(key); block == nil {
+		return nil, ErrKeyMustBePEMEncoded
+	}
+	var asn1PubKey ed25519PubKey
+	if _, err := asn1.Unmarshal(block.Bytes, &asn1PubKey); err != nil {
```
oxisto, line 48: See comment above.

vtolstov:
done

vtolstov (commented):
anything else must be fixed before merge?

oxisto:
> anything else must be fixed before merge?

Looks good overall, I will check the merge tomorrow.

ed25519.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"crypto/ed25519"
+)
+
+var (
+	ErrEd25519Verification = errors.New("ed25519: verification error")
+)
+
+// Implements the EdDSA family
+// Expects *ed25519.PrivateKey for signing and *ed25519.PublicKey for verification
```
oxisto, line 14: Minor nit pick: it actually does expect `ed25519.PrivateKey` rather than `*ed25519.PrivateKey`, if you look at the `Sign` method below.  But this seems to be the expected outcome of calling `ParsePKCS8PrivateKey`. In the ecdsa case, it contains a `*ecdsa.PublicKey`.

 I wonder, if this is an inconsistency in the Go crypto API or is `ed25519.PrivateKey` an interface rather than a struct?

vtolstov: don't know, but i'm fix comments =)

oxisto (approved):

oxisto:
LGTM, I will give the others, @mfridman @Waterdrips @lggomez some time to object, as we agreed on having two approvals ideally. Otherwise, we can merge this in over the weekend.

vtolstov:
also will be fine to get new release after this will be merged (to get allocations optimization and eddsa support)

ripienaar (approved):
LGTM

PR #40, opened 2021-07-29
upstream fix for security vulnerability from form3tech-oss/jwt-go fork

thaJeztah:
This forwards the changes of https://github.com/form3tech-oss/jwt-go/pull/14 to the upstream repository.

I am not the author of the original PR (nor do I know much about JWT), but thought I'd give this an attempt, following the discussion in https://github.com/Azure/go-autorest/issues/642 (and issues linked from there).

Fixes a security vulnerability where a jwt token could potentially be validated having invalid string characters.

(cherry picked from commit a211650c6ae1cff6d7347d3e24070d65dcfb1122)
https://github.com/form3tech-oss/jwt-go/pull/14

Issue #30, opened 2021-06-22
Proper go.mod support


harshavardhana:
Such that moving the API to `github.com/golang-jwt/jwt/v3`  for v3 releases. This helps downstream projects such as github.com/minio/minio - to pin to a correct release.

The problem with `go.mod`'s - `v3.2.1+incompatible` would simply be unexpected

The original issue that was neverhttps://github.com/dgrijalva/jwt-go/issues/294

oxisto:
HI @harshavardhana, thanks for the interest in this project. I am not quite sure if I understand your issue correctly. The `+ incompatible` is just an indication of Go's forced SIV-style being "ignored", because the original author did so. It should not hinder you in any way to "pin" to a certain release, such as the current `v3.2.1+incompatible`.

We had a rather long going-forth-and-back-discussion on this topic, mainly in #17 and #5 and decided that we do want to support go modules properly at some point, but not during the `v3` releases. This is mainly due to the fact that we wanted to have a 1:1 drop-in replacement first, fixing major things and then move to a `v4` release. We decided against switching to a SIV-style import path during the `v3` releases and moved this to `v4`.

We have yet to fix a proper release schedule, I have started something here: https://github.com/golang-jwt/jwt/milestones, but

harshavardhana:
> HI @harshavardhana, thanks for the interest in this project. I am not quite sure if I understand your issue correctly. The `+ incompatible` is just an indication of Go's forced SIV-style being "ignored", because the original author did so. It should not hinder you in any way to "pin" to a certain release, such as the current `v3.2.1+incompatible`.
> 
> We had a rather long going-forth-and-back-discussion on this topic, mainly in #17 and #5 and decided that we do want to support go modules properly at some point, but not during the `v3` releases. This is mainly due to the fact that we wanted to have a 1:1 drop-in replacement first, fixing major things and then move to a `v4` release. We decided against switching to a SIV-style import path during the `v3` releases and moved this to `v4`.
> 
> We have yet to fix a proper release schedule, I have started something here: https://github.com/golang-jwt/jwt/milestones, but 

Yes I saw that after opening the issue, looks like there is no easy way out. Looking forward to v4 implementation.

oxisto:
I will keep this open, until we address it in v4.

mfridman:
One could make the case this package doesn't _really_ need module support. And the only real side-effect afaics is the "ugly" `+incompatible` suffix

1. there are no external dependencies
[...]

Issue #32, opened 2021-06-30
Incorporate string array in `aud` field from the original v4 branch


dcormier:
The original repo has [a pretty complete v4 release](https://github.com/dgrijalva/jwt-go/tree/release_4_0_0) that I'm using ([docs](https://pkg.go.dev/github.com/dgrijalva/jwt-go/v4)).

As I recall, I needed v4 because of [this change](https://github.com/dgrijalva/jwt-go/blob/release_4_0_0/VERSION_HISTORY.md#user-content-400):

> Added new type, `ClaimStrings`, which will correctly handle properties such as `aud` that can be either an array of strings or a single string. `ClaimStrings` is an alias to `[]string`, but with custom decoding behavior. This means all the built in `aud` validation behavior now expects `[]string` instead of `string`. This was a common request.

It would be nice to see that picked up and moved forward, as well.

oxisto:
I have tried to put that functionality in https://github.com/golang-jwt/jwt/pull/15. Could you have a look whether this fulfills your use case?

dcormier:
It looks like that probably would handle that my use case.

dunglas:
This is currently a BC break. When using `StandardClaims` with an `aud` field containing a string, this error is now thrown: `unable to parse JWT: json: cannot unmarshal array into Go struct field claims.aud of type string`.

Using the new `RegisteredClaims` introduced in https://github.com/golang-jwt/jwt/pull/15 works properly, but even with this patch the BC break is still present of the deprecated `StandardClaims` struct.

PR #41, opened 2021-07-29
Adds go module support /v4

### LINKED ISSUE #5: Add go module support
The existing project https://github.com/dgrijalva/jwt-go had no module support and was likely imported as:

```
github.com/dgrijalva/jwt-go v3.2.0+incompatible
```

There was also a [v4](https://github.com/dgrijalva/jwt-go/releases/tag/v4.0.0-preview1) that never got fully released.

```
go get github.com/dgrijalva/jwt-go@v4.0.0-preview1
go get: github.com/dgrijalva/jwt-go@v4.0.0-preview1: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v4
```

---

Since users would import a new module anyways, one solution could be drop _all_ existing versions (14 total) and start with a fresh `v1.0.0` (hopefully no breaking changes required).

For many of my projects I see this being a basic search/replace operation:

```diff
- github.com/dgrijalva/jwt-go v3.2.0+incompatible
+ github.com/golang-jwt/jwt v1.0.0
```

The version in this case is irrelevant afaics.

---

Lastly, I am not sure whether the original repo will get "transferred" or "archived" ([see comment here](https://github.com/dgrijalva/jwt-go/issues/462#issuecomment-846423835)), but if it gets transferred then this might change the outcome of how to proceed adding module support?

Thoughts, suggestions welcome.

mfridman:
Introduces modules support with with`/v4` SIV

- [x] Update migration guide
- [x] Update README to ensure import paths are correct referenced

---

fixes #5 #3 #30


oxisto:
To be completely honest, I do not see a big advantage of doing this now. There was an initial wave of people coming over from  the old repo to this one now because of the fact that GitHub introduced dependency / security updates for Go, so they were triggered by the dependant security alert of the CVE that was fixed in 3.2.1. Yes of course there were some issues, especially with transitive dependencies, but that was to be expected. I would suggest to let this be settled, take a few weeks to give also major projects to move the dependency over to this project for the 3.2.1. Then we need to fix #40, which could be considered a security fix in some undesired configurations and give people a chance to easily move from 3.2.1 to 3.2.2.

IMHO, creating a v4 release _now_ would just add more chaos instead of actually doing any good. But this is just my 2 cents.

ripienaar:
I am inclined to agree. Adding a go mod does not require v4. Though I would love to see the go mod added

thaJeztah:
adding a go.mod without changing the import path _is_ a breaking change (please don't!) without a go.mod, modules will allow the module (with `+incompatible`), but after a `go.mod` is added, it will refuse to use it.

ripienaar:
Ah you are right sorry. I forgot the version was already at 3. Then yes we need to bump the version but vote then to wait to get all the inflight stuff done at the same time before tagging 4

thaJeztah:
> Then yes we need to bump the version but vote then to wait to get all the inflight stuff done at the same time before tagging 4

Yes, I think that would make sense;

1. tag a "final" v3.x release (without `go.mod`)
2. optionally: create a `v3` branch from last release, to allow backporting security fixes to v3 for some time
3. move main / master branch to v4, and add `go.mod`
4. tag `v4.0.0` (no changes in functionality with v3)

Step `4.` in the above allows users to migrate from v3 (no modules) to v4 (modules) without risk of pulling in other changes, which reduces friction for users to migrate to v4 (and thus (hopefully) the need for having to maintain the v3 branch for an extended period of time)

It _would_ mean though that if there are plans to introduce breaking changes, a `v5.0.0` release would be needed after that, so

`5.` introduce breaking changes
`6.` rename module in main branch to `github.com/golang-jwt/jwt/v5`
`7.` release as `v5.0.0`

mfridman:
> To be completely honest, I do not see a big advantage of doing this now. There was an initial wave of people coming over from the old repo to this one now because of the fact that GitHub introduced dependency / security updates for Go, so they were triggered by the dependant security alert of the CVE that was fixed in 3.2.1. Yes of course there were some issues, especially with transitive dependencies, but that was to be expected. I would suggest to let this be settled, take a few weeks to give also major projects to move the dependency over to this project for the 3.2.1. Then we need to fix #40, which could be considered a security fix in some undesired configurations and give people a chance to easily move from 3.2.1 to 3.2.2.
> 
> IMHO, creating a v4 release _now_ would just add more chaos instead of actually doing any good. But this is just my 2 cents.

I agree with some of these points, but waiting it just fighting the inventible. Also the earlier we do it, the fewer folks will have to migrate (if they so choose).

It's unfortunate that we're _forced_ into a SIV style import, but this is the landscape of Go modules. My 2 cents is the sooner we can adopt modules the less painful it will be.

---

@thaJeztah wrt to point 3:

> 3. move main / master branch to v4, and add go.mod

Suppose we merge this PR, and then tag a v4.0.0 on `main`.. is the concern that users on `v3.x.y+incompatible` will be broken?

If I understood you correctly, you're suggesting we add/maintain `v3` (non modules) and `v4` as default branch (modules) branches while backporting fixes?

thaJeztah:
> Suppose we merge this PR, and then tag a v4.0.0 on main.. is the concern that users on v3.x.y+incompatible will be broken?

No `v3.x.y+incompatible` will continue to be working, as long as any `v3.x.y+incompatible` tag does not contain a `go.mod`

> If I understood you correctly, you're suggesting we add/maintain v3 (non modules) and v4 as default branch (modules) branches while backporting fixes?

I would consider `v3` to be marked "deprecated" once `v4` is released, and _hope_ no backporting will be needed, except for the rare exception where a critical vulnerability is found before consumers of this module have been able to migrate to v4.
Reason for _considering_ to create a `v3` branch is "allow" that exceptional situation to occur, and to prevent "yet another fork" to come to existence. Perhaps it would be ok to clearly mention this in the README; containerd did a good job at that; https://github.com/containerd/containerd/blob/4fdb884644ecdd6a4b4b0d1bd85b8f6ccc1995c6/RELEASES.md#support-horizon
The README could mention that the v3 version, and v3 branch should be considered EOL (or "extended support"), and only critical security patches will be accepted for backporting to that branch (any other backport must be rejected).

> It's unfortunate that we're forced into a SIV style import, but this is the landscape of Go modules. My 2 cents is the sooner we can adopt modules the less painful it will be.

I have quite a love/hate relationship with go modules, but in general agree that it would be good to move to modules (v4) as soon as possible.

Under "regular" circumstances, I would say; merge and move, but if https://github.com/golang-jwt/jwt/pull/40 is indeed a security fix, I would highly recommend doing the move to v4 _after_ that is merged.

Security releases should have as little friction as possible, and requiring users to accept a breaking change (move to v4, and because of that incompatible with all existing uses of the module), definitely will cause considerable more friction, so if it's possible to postpone that move until after https://github.com/golang-jwt/jwt/pull/40 is merged, please do so (again, assuming that's a security fix that users may need to have).

Looking at this PR w.r.t. "pros/cons" to move to modules "ASAP";

- :+1:/:-1: This project has _no_ external dependencies, so from that perspective, moving to go modules won't bring benefits to users (modules help to resolve what versions of other dependencies are needed, in this case that doesn't apply)
- :+1:/:-1: No more `+incompatible` in user's `go.mod`. Nice to have, but other than being a bit of a nuisance to see the `+incompatible`, no "urgent" fix
- :-1: Moving to v4 is a breaking change; this is ok, but (ideally) shouldn't prevent users from being able to get a security fix (as outlined above).
- :+1: Moving to modules "v4" opens up the opportunity to accept breaking changes ("v5"). Despite sometimes being a "pain", the versioned import paths allows users to have multiple versions of the module if needed (they can have both v3 (pre-modules), v4, and v5) in the same codebase.

oxisto:
@mfridman Good points indeed. Ok then let's release 3.2.2; #40 was just merged in and then let's bite the bullet and do a v4 release immediately after that.

mfridman:
Thanks for the detailed response. 

One more 👍 I'd give is to solve transitive dependancy issues esp. when using `replace` directives (which should not be needed). At that point users should drop `replace` and instead switch to the `/v4`, effectively treating this as a separate module altogether.

I pushed a `v3` branch https://github.com/golang-jwt/jwt/tree/v3 .. branched off from current `main` 4bbdd8a which is tagged as `v3.2.2`. This contains the fix https://github.com/golang-jwt/jwt/pull/40

thaJeztah:
Thank you so much! Have a great weekend 👍

oxisto (dismissed):
I am still not completely happy with this (not with the PR, but with the whole situation), but I approve :)

oxisto:
@mfridman Could you please remove #32 from the list of linked issues (I somehow lack the permissions to do so), because this issue relates to the missing multiple strings in `aud`, which is going to be fixed in #15, but not in this PR.

oxisto (approved):
Let's do it.

mfridman:
@oxisto I won't be able to prepare a release for a few hours (meetings/day job). Any chance yourself or one of the other maintainers can prepare up a `v4.0.0` ?

PR #44, opened 2021-07-31
Added `staticcheck` for basic static code analysis

oxisto:
I am basing this off the `v4` branch because getting this to work without Go modules is a pain.

Addressed all the issues that static check discovered. Also deprecated several functions that were just used internally.


cmd/jwt/main.go
```diff
@@ -67,14 +67,14 @@ func start() error {
 		return showToken()
 	} else {
 		flag.Usage()
-		return fmt.Errorf("None of the required flags are present.  What do you want me to do?")
+		return fmt.Errorf("none of the required flags are present.  What do you want me to do?")
```
lggomez, line 70: Can we consider this a non-breaking change? personally I fully agree with correcting this style but this will break user tests that perform assertions against the explicit errors values. Considering this goes for v4, we can consider it a non-issue but I would think twice if we ever need to backport to v3

lggomez (approved):
LGTM (the above comment can be a separate thread)

cmd/jwt/main.go
oxisto: Since it is in the actual `jwt` application, I would say that this is ok. For the other error changes I would assume that people are trying to compare against the error variables and not the content. But anyway, yes, I would keep this for v4, which promises compatibility but considering it to be a major version bump I would argue this has enough leeway to fix issues like this.

Issue #45, opened 2021-07-31
x509.DecryptPEMBlock has been deprecated since Go 1.16


oxisto:
We should remove/replace it, to be compatible with future Go versions.

oxisto:
It looks like there is no alternative in the Go standard library, so we are somewhat stuck here. Not sure how many people actually use `ParseRSAPrivateKeyFromPEMWithPassword`, it is in the utils file and primarily used for a test. We should at least deprecate it and remove it in a future version, i.e. v4 or v5.

leo-at-itartpro:
I use it, it was actually in a Udemy course I saw last year. What's the alternative? Omit the password altogether?

HimanshuSharma-Angel:
Any alternative library?

oxisto:
> Any alternative library?

I am afraid we don’t want the hassle of including anything else but the standard library. Especially for something which is not really part of the “public” API.

PR #46, opened 2021-07-31
Deprecate legacy PEM encryption

oxisto:
Deprecating `ParseRSAPrivateKeyFromPEMWithPassword` since the underyling `x509.DecryptPEMBlock` is deprecated in Go 1.16, since RFC 1423 is considered to be insecure by design.

Related to #45, but does not completely fix it.


mfridman (approved):

oxisto:
This was already done as part of the v4 release. Closing this.

PR #51, opened 2021-08-01
Add back support for Go 1.14

### LINKED ISSUE #50: Due to the use of fillBytes golang-jwt now requires go1.15
https://github.com/golang-jwt/jwt/blob/4bbdd8ac624fc7a9ef7aec841c43d99b5fe65a29/ecdsa.go#L135-L136

from #33, means that golang-jwt/jwt now requires go1.15.

Is this intentional?

zeripath:
Adds simple reimplementation of fillBytesInt for go 1.14 support

Fix #50

Signed-off-by: Andrew Thornton <art27@cantab.net>

### DIFF
--- a/ecdsa_go1.14.go
+++ b/ecdsa_go1.14.go
@@ -60,15 +60,24 @@ func fillBytesInt(x *big.Int, buf []byte) []byte {
 	for i := range buf {
 		buf[i] = 0
 	}
+	// The following is mostly copied from the original go source code - however it is the only way of doing this.
+
+	// Start at the end of the buffer as x.Bits() is a little endian and work for wards
 	i := len(buf)
+
+	// Walk the words from x.Bits()
 	for _, d := range x.Bits() {
+		// Now each word is Uintsize (usually 64 bits) in length - but a byte is 8 bits so each byte must be split in Uintsize/8 slices
 		for j := 0; j < (bits.UintSize / 8); j++ {
+			// Move forward one step
 			i--
 			if i >= 0 {
+				// set the value of the current buf[i] to the byte value of
 				buf[i] = byte(d)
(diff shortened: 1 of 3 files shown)


mfridman:
Can you please add go1.14 to workflow?

https://github.com/golang-jwt/jwt/blob/main/.github/workflows/build.yml#L16

mfridman:
Ah perfect, you added build tags

ecdsa_go1.14.go
```diff
@@ -0,0 +1,77 @@
...
+		buf[i] = 0
+	}
+	i := len(buf)
+	for _, d := range x.Bits() {
+		for j := 0; j < (bits.UintSize / 8); j++ {
+			i--
+			if i >= 0 {
+				buf[i] = byte(d)
+			} else if byte(d) != 0 {
+				panic("math/big: buffer too small to fit value")
```
oxisto, line 70: For me this looks like a 1:1 copy out of the Go std lib, even including the particular error here and the comment above. Not quite sure, if this ok from a license point of view. Even though it is BSD, it still needs proper attribution and unless you are the original author of that line, this is very fishy in my opinion to just copy/paste that here.

oxisto (changes requested):

ecdsa_go1.14.go
zeripath: Adjusted.

mfridman: I had the exact same thought (was just about to check if this was a copy/paste from the standard library).

I usually reference this ticket by Ian. https://github.com/golang/go/issues/19893#issuecomment-292755517

zeripath: The panic has to be exactly the same otherwise it's not compatible.

PR #52, opened 2021-08-03
Add a Go module file

xyproto:


oxisto:
We are working on providing Go modules in version v4 through PR #41. We cannot switch to Go modules with the current import path.

Closing as a duplicate of #41

PR #53, opened 2021-08-03
Format code with "go fmt"

xyproto:


lggomez (approved):

lggomez:
LGTM. @mfridman @oxisto 

As a further pass I propose using https://github.com/mvdan/gofumpt, which is a superset of go fmt maintained by some of the folks of the go team (my guess is some of the rules eventually make into fmt, but this is the bleeding edge version)

mfridman:
LGTM.

I like gofumt, but _most_ users are unlikely to have this enabled. E.g., if using vs code requires this setting:

```
"gopls": {
    "formatting.gofumpt": true
}
```

Were you thinking of adding a github workflow that automatically formats code after each merge to `main` ?

PR #55, opened 2021-08-03
Remove init functions by adding functions for creating new SigningMet…

xyproto:
* Add new functions for creating SigningMethod structs, like `NewSigningMethodECDSA`.
* Use `var` instead of `init` functions.
* Let RegisterSigningMethod take a `SigningMethod` type, and let `RegisterSigningMethodFunc` take an algorithm name and a function that returns a SigningMethod.
* The suggested changes adds 122 lines and removes 147 lines.
* All tests pass.

### DIFF
--- a/ecdsa.go
+++ b/ecdsa.go
@@ -30,7 +30,7 @@ var (
 )
 
 // NewSigningMethodECDSA creates a new SigningMethodECDSA struct and
-// registers it with RegisterSigningMethod.
+// registers it as a signing method.
 func NewSigningMethodECDSA(name string, hash crypto.Hash, keySize, curveBits int) *SigningMethodECDSA {
 	m := &SigningMethodECDSA{name, hash, keySize, curveBits}
 	Register(m)
--- a/ecdsa.go
+++ b/ecdsa.go
@@ -24,29 +24,17 @@ type SigningMethodECDSA struct {
 
 // Specific instances for EC256 and company
 var (
-	SigningMethodES256 *SigningMethodECDSA
-	SigningMethodES384 *SigningMethodECDSA
-	SigningMethodES512 *SigningMethodECDSA
+	SigningMethodES256 = NewSigningMethodECDSA("ES256", crypto.SHA256, 32, 256)
+	SigningMethodES384 = NewSigningMethodECDSA("ES384", crypto.SHA384, 48, 384)
+	SigningMethodES512 = NewSigningMethodECDSA("ES512", crypto.SHA512, 66, 521)
 )
 
-func init() {
-	// ES256
-	SigningMethodES256 = &SigningMethodECDSA{"ES256", crypto.SHA256, 32, 256}
-	RegisterSigningMethod(SigningMethodES256.Alg(), func() SigningMethod {
-		return SigningMethodES256
-	})
-
-	// ES384
-	SigningMethodES384 = &SigningMethodECDSA{"ES384", crypto.SHA384, 48, 384}
-	RegisterSigningMethod(SigningMethodES384.Alg(), func() SigningMethod {
-		return SigningMethodES384
-	})
-
-	// ES512
-	SigningMethodES512 = &SigningMethodECDSA{"ES512", crypto.SHA512, 66, 521}
-	RegisterSigningMethod(SigningMethodES512.Alg(), func() SigningMethod {
-		return SigningMethodES512
-	})
+// NewSigningMethodECDSA creates a new SigningMethodECDSA struct and
+// registers it with RegisterSigningMethod.
+func NewSigningMethodECDSA(name string, hash crypto.Hash, keySize, curveBits int) *SigningMethodECDSA {
+	m := &SigningMethodECDSA{name, hash, keySize, curveBits}
+	Register(m)
+	return m
 }
 
 func (m *SigningMethodECDSA) Alg() string {
@@ -56,11 +44,10 @@ func (m *SigningMethodECDSA) Alg() string {
 // Verify implements token verification for the SigningMethod.
 // For this verify method, key must be an ecdsa.PublicKey struct
 func (m *SigningMethodECDSA) Verify(signingString, signature string, key interface{}) error {
-	var err error
 
 	// Decode the signature
-	var sig []byte
-	if sig, err = DecodeSegment(signature); err != nil {
+	sig, err := DecodeSegment(signature)
+	if err != nil {
 		return err
 	}
 
@@ -96,7 +83,7 @@ func (m *SigningMethodECDSA) Verify(signingString, signature string, key interfa
 }
 
 // Sign implements token signing for the SigningMethod.
-// For this signing method, key must be an ecdsa.PrivateKey struct
+// For this signing method, the key must be an ecdsa.PrivateKey struct.
 func (m *SigningMethodECDSA) Sign(signingString string, key interface{}) (string, error) {
 	// Get the key
 	var ecdsaKey *ecdsa.PrivateKey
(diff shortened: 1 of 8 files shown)


xyproto:
The two commits should be squashed. The commit message for the "go fmt" commit is not correct.

oxisto:
I need some time to look at this properly. It does change the public API though.

xyproto:
I updated the pull request so that it does not change the API.

BryceDFisher (dismissed):
I like moving it to a more interface centric approach.  It looks much more transportable for new signing methods in the future.

map_claims_test.go
```diff
@@ -7,55 +7,52 @@ import (
 func TestVerifyAud(t *testing.T) {
 	var nilInterface interface{}
 	var nilListInterface []interface{}
-	var intListInterface interface{} = []int{1,2,3}
```
lggomez, line 10: The changes on this file are only from go fmt? if that's the case I'd revert them the there is no change in the scope of this PR

lggomez (changes requested):

lggomez:
Except for the format-only file change it LGTM. While changing the API it is only extending it and it improves the package initialization workflow (since init methods are called by the lexicographical order of the files they're in, making them flaky if one package contains several files with init methods sharing logic or state)

map_claims_test.go
xyproto: Thanks for reviewing! I'll make the requested changes.

xyproto: I removed changes related to map_claims_test.go.

lggomez: I still see the file changed, could it be an editor setting changing it automatically?

rsa.go
```diff
@@ -15,29 +15,17 @@ type SigningMethodRSA struct {
...
-	})
-
-	// RS512
-	SigningMethodRS512 = &SigningMethodRSA{"RS512", crypto.SHA512}
-	RegisterSigningMethod(SigningMethodRS512.Alg(), func() SigningMethod {
-		return SigningMethodRS512
-	})
+// NewSigningMethodRSA creates a new SigningMethodRSA struct and
+// registers it as a signing method.
+func NewSigningMethodRSA(name string, hash crypto.Hash) *SigningMethodRSA {
```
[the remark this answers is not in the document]
xyproto: I can change the functions so that they start with `newSigningMethod` instead.

signing_method.go
```diff
@@ -14,22 +14,30 @@ type SigningMethod interface {
 	Alg() string                                                   // returns the alg identifier for this method (example: 'HS256')
 }
 
-// RegisterSigningMethod registers the "alg" name and a factory function for signing method.
-// This is typically done during init() in the method's implementation
-func RegisterSigningMethod(alg string, f func() SigningMethod) {
+// Register create a new function that returns the given
```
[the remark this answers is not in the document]
xyproto: It does create a new function, with `func() SigningMethod { return m }`?

rsa.go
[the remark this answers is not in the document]
xyproto: I updated the pull request. :heavy_check_mark:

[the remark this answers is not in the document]
lggomez: LIked this, was a good afterthought. Now I think it should be ready to go, it LGTM from my side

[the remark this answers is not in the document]
xyproto: @lggomez GitHub believes there is one outstanding request for change.

Could you please take a look if this is still the case, because I can't spot it?

lggomez (approved):
LGTM, moved the changed request to approved from my side. Please take a final check to confirm @oxisto

ecdsa.go
```diff
@@ -115,28 +102,29 @@ func (m *SigningMethodECDSA) Sign(signingString string, key interface{}) (string
...
 
-	// Sign the string and return r, s
-	if r, s, err := ecdsa.Sign(rand.Reader, ecdsaKey, hasher.Sum(nil)); err == nil {
-		curveBits := ecdsaKey.Curve.Params().BitSize
-
-		if m.CurveBits != curveBits {
-			return "", ErrInvalidKey
-		}
+	// Sign the string
+	r, s, err := ecdsa.Sign(rand.Reader, ecdsaKey, hasher.Sum(nil))
```
oxisto, line 106: Personally, I am not a big fan of changing the `Sign` and `Verify` functions within this PR, even if this is most likely purely cosmetical.

signing_method.go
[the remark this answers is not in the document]
oxisto:
```suggestion
// Register internally creates a new function that returns the given
```

oxisto (approved):
I generally approve. One minor documentation nitpick and personally I do not like changing the `Sign` and `Verify` methods as part of this PR - even through I agree that they are more readable this way.

ecdsa.go
xyproto: Yes, it's just deindentation to make the function more readable, as suggested by go vet / golint / staticcheck.

signing_method.go
```diff
@@ -14,22 +14,30 @@ type SigningMethod interface {
 	Alg() string                                                   // returns the alg identifier for this method (example: 'HS256')
 }
 
-// RegisterSigningMethod registers the "alg" name and a factory function for signing method.
-// This is typically done during init() in the method's implementation
-func RegisterSigningMethod(alg string, f func() SigningMethod) {
+// Register internally creates a new function that returns the given
+// SigningMethod. The function will be stored as a value in a thread safe map,
+// where the algorithm name is the key.
+func Register(m SigningMethod) {
```
oxisto, line 20: Just another small nitpick. Can we have a different name than just `Register`? As we do not have any sub packages, having a `Register` function directly at the public package level might be confusing.

Also, is it intended to have this exported? I guess to support external packages? Or should we direct them to `RegisterSigningMethod` (if this is actually still working now?). In this case, we might decide to make it private.

mfridman: I'd vote to keep `Register` unexported.  Consumers already have access to `RegisterSigningMethod` which does the same thing, albeit in a less friendly way.

But I think having a smaller API surface and a single way is better overall.

xyproto: I would prefer if the `Register(m SigningMethod)` function was named `RegisterSigningMethod(m SigningMethod)`, but `RegisterSigningMethod` is already used for `RegisterSigningMethod(alg string, f func() SigningMethod)`, and renaming that function or making it lowercase would change the API.

I think it would be helpful if `Register(m SigningMethod)` was exported, because it's a nice way to register new `SigningMethod`s with `golang-jwt/jwt`. Used from another package, it would have the context of `jwt`, like this: `jwt.Register(mySigningMethod)`, so it would not be without context.

I will happily change this pull request if there are suggestions for a better name than `Register` (or if it would be possible to change the API so that the exported function would be `RegisterSigningMethod(m SigningMethod)` instead of `RegisterSigningMethod(alg string, f func() SigningMethod)`. I feel like just unexporting `Register` by renaming it to `register` is a bit tame solution, and would not provide the best API for users.

mfridman: I think the name is fine, and what you've proposed absolutely makes sense. 

The issue is adding yet another way to do something. As a consumer I know have to understand the difference between `RegisterSigningMethod` vs `Register` and that's already too much work.

I agree this package has a lot of room for improvement, but if we continue duplicating existing functions / methods then it becomes confusing for both old and new users, the former needs to understand if there is an actual improvement in the newly added API  and new users need to understand the difference between the two.

I think it would make sense to replace `RegisterSigningMethod` with `Register`  in a `/v5` release.

The alternative is to add a `Register` as you proposed with a deprecation notice to `RegisterSigningMethod`.

xyproto: Thanks for the insightful response, @mfridman. I added a deprecation comment to the pull request.

I also believe that replacing `RegisterSigningMethod` with the `Register` function would make sense for a `v5` release.

ecdsa.go
xyproto: I assume that jwt wishes to follow best practices and the advice of go vet / golint / staticcheck, as long as it does not hinder readability.

Please let me know if this should rather be in a separate pull request.

signing_method.go
mfridman: @oxisto and co. Did you have further opinions on `RegisterSigningMethod` and `Register` ? I know we're bideshedding here a bit, but it's worth discussing as it sets a precedent for the package as a whole.

I still worry about the number of PRs introducing new exportable functions. Albeit adding value, they do introduce complexity and duplication into the codebase.

xyproto: Alternatively, the `RegisterSigningMethod` and `Register` method can be combined into a single `RegisterSigningMethod` by using duck typed arguments.

This way the API is kept stable, type checking is still in place, it is possible to just pass in a single struct that implements the SigningMethod interface, and all tests pass here:

```go
func RegisterSigningMethod(algOrSigningMethods ...interface{}) error {
	switch len(algOrSigningMethods) {
	case 1:
		// One argument is given, check if it is a SigningMethod
		m, ok := algOrSigningMethods[0].(SigningMethod)
		if !ok {
			return errors.New("one argument is given, and it does not implement SigningMethod")
		}
		signingMethodLock.Lock()
		defer signingMethodLock.Unlock()
		signingMethods[m.Alg()] = func() SigningMethod { return m }
		return nil
	case 2:
		// Two arguments are given, check if it is a string and a func() SigningMethod
		alg, ok := algOrSigningMethods[0].(string)
		if !ok {
			return errors.New("two arguments are given and the first argument is not a string")
		}
		f, ok := algOrSigningMethods[1].(func() SigningMethod)
		if !ok {
			return errors.New("two arguments are given and the second argument is not a function that returns a SigningMethod")
		}
		signingMethodLock.Lock()
		defer signingMethodLock.Unlock()
		signingMethods[alg] = f
		return nil
	}
	return errors.New("too few or too many arguments are given")
}
```

xyproto: @mfriedman:

@oxisto and others does not appear to have further opinions on this

lggomez:
LGTM. @oxisto do you think we can have this merged fox 4.x? It does modify the API indeed but the code refactor is also one of the main points of this stale PR

mfridman:
I think this PR went stale, going to close this for now.

xyproto:
What can I do to un-stale it?

Rebase it on main and/or re-submit the PR?

PR #84, opened 2021-08-03
jwt: Return better error messages on malformed tokens

### LINKED ISSUE #75: jwt.Parse got "invalid character 'x' looking for beginning of value"
Migrated from https://github.com/dgrijalva/jwt-go/issues/361:

>  alanywlee commented on Oct 8, 2019
> 
> I am trying to parse a RS256 encoded token (tokenString) from java, it could be successfully decoded by public key in java using the following java code
> 
> Jwts.parser().setSigningKey(publicKey).parseClaimsJws(jwt);
> 
> However, It could be verified by jwt.SigningMethodRS256.Verify() just as described in https://stackoverflow.com/questions/51834234/i-have-a-public-key-and-a-jwt-how-do-i-check-if-its-valid-in-go
> 
> But not able to be jwt.Parse, I tried to write more debug print messages inside the parse function, and verified not even run into it.

```go
    tokenString := "eyJhbGciOiJSUzI1NiIsInppcCI6IkRFRiJ9.eNqqVkqtKFCyMjQ1s7Q0sbA0MtFRyk3NTUot8kxRslIKLbZQggn4JeamAoUcfRz99HxcXRWeze172tr4bFq7Ui0AAAD__w.jBXD4LT4aq4oXTgDoPkiV6n4QdSZPZI1Z4J8MWQC42aHK0oXwcovEU06dVbtB81TF-2byuu0-qi8J0GUttODT67k6gCl6DV_iuCOV7gczwTcvKslotUvXzoJ2wa0QuujnjxLEE50r0p6k0tsv_9OIFSUZzDksJFYNPlJH2eFG55DROx4TsOz98az37SujZi9GGbTc9SLgzFHPrHMrovRZ5qLC_w4JrdtsLzBBI11OQJgRYwV8fQf4O8IsMkHtetjkN7dKgUkJtRarNWOk76rpTPppLypiLU4_J0-wrElLMh1TzUVZW6Fz2cDHDDBACJgMmKQ2pOFEDK_vYZN74dLCF5GiTZV6DbXhNxO7lqT7JUN4a3p2z96G7WNRjblf2qZeuYdQvkIsiK-rCbSIE836XeY5gaBgkOzuEvzl_tMrpRmb5Oox1ibOfVT2KBh9Lvqsb1XbQjCio2CLE2ViCLqoe0AaRqlUyrk3n8BIG-r0IW4dcw96CEryEMIjsjVp9mtPXamJzf391kt8Rf3iRBqwv3zP7Plg1ResXbmsFUgOflAUPcYmfLug4W3W52ntcUlTHAKXrNfaJL9QQiYAaDukG-ZHDytsOWTuuXw7lVxjt-XYi1VbRAIjh1aIYSELEmEpE4Ny74htQtywYXMQNfJpB0nNn8IiWakgcYYMJ0TmKM"

    token, err := jwt.Parse(tokenString, func(token *jwt.Token) (interface{}, error) {
        fmt.Println("token parse start...")
    ...
    }
    // ... error handling
    if err != nil {
        log.Fatal(err)
    }
```

> It got error message as following
> 
> 2019/10/08 17:29:38 invalid character 'x' looking for beginning of value
> exit status 1
> 
> I am curious if I can verify token signature, how come not able to parse its header and claims body. Or I missed anything before doing jwt.Parse()
> 
> Any comment is highly appreciated....

lggomez:
Fixes #75, as the jwt.io debugger also considers this payload invalid JSON and is validated as expected

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -107,10 +107,10 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		if strings.HasPrefix(strings.ToLower(tokenString), "bearer ") {
 			return token, parts, NewValidationError("tokenstring should not contain 'bearer '", ValidationErrorMalformed)
 		}
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token header: %w", err), Errors: ValidationErrorMalformed}
 	}
 	if err = json.Unmarshal(headerBytes, &token.Header); err != nil {
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token header: %w", err), Errors: ValidationErrorMalformed}
 	}
 
 	// parse Claims
@@ -118,7 +118,7 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	token.Claims = claims
 
 	if claimBytes, err = DecodeSegment(parts[1]); err != nil {
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token claims: %w", err), Errors: ValidationErrorMalformed}
 	}
 	dec := json.NewDecoder(bytes.NewBuffer(claimBytes))
 	if p.UseJSONNumber {
@@ -132,7 +132,7 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	}
 	// Handle decode error
 	if err != nil {
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token payload: %w", err), Errors: ValidationErrorMalformed}
 	}
 
 	// Lookup signature method
--- a/parser_test.go
+++ b/parser_test.go
@@ -287,6 +287,21 @@ func TestParser_ParseUnverified(t *testing.T) {
 	}
 }
 
+func Test_ParseWithInvalidJSONPayload(t *testing.T) {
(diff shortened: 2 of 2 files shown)


parser.go
```diff
@@ -107,18 +107,18 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		if strings.HasPrefix(strings.ToLower(tokenString), "bearer ") {
 			return token, parts, NewValidationError("tokenstring should not contain 'bearer '", ValidationErrorMalformed)
 		}
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token header: %w", err), Errors: ValidationErrorMalformed}
 	}
 	if err = json.Unmarshal(headerBytes, &token.Header); err != nil {
-		return token, parts, &ValidationError{Inner: err, Errors: ValidationErrorMalformed}
+		return token, parts, &ValidationError{Inner: fmt.Errorf("malformed token header: %w", err), Errors: ValidationErrorMalformed}
```
mfridman, line 113: I've been going back and forth on this one. The intention is good, but is it going to break existing users?

There is _a lot_ to be desired in terms of improving errors. But is this something we should tackle in the current `/v4` or punt to a breaking release `/v5`.

oxisto: This discussion comes back to the point that we should discuss a common roadmap. I wonder, whether doing this in GitHub discussions is enough or if it makes sense to have a short zoom / Teams call on this to discuss a some roadmap ideas among the maintainers. Probably tricky to find a good time slot for this.

lggomez: We can opt to use semver properly in v4 and treat minor but non-revisional changes as these via releases (unless we strictly want to make the v4 release family v3 compatible). Those are the 2 options I see right now

@oxisto I think we can do fine with gh discussions for a middle ground between a chat and a an issue, for something more fluent we could use discord or something alike. I agree that coordinating a time slot might be difficult for a live conversation as we have tight schedules and different timezones but a chat server comes second place in communication efficiency so we can still gain something

oxisto: Yeah, maybe a discord server would be a nice idea.

mfridman: > We can opt to use semver properly in v4 and treat minor but non-revisional changes as these via releases

Ye that's fair, since we are using modules we can bump minor versions and just make sure to jot that down in the release change log.

I'm not sure I have the bandwidth for in-person chats. Happy to continue contributing asynchronously, also there is nothing pressing afacis. 

FWIW there is already a Gophers slack, and can create a "jwt" channel (maybe not) or just ping whoever is on there.

parser_test.go
```diff
@@ -287,6 +287,21 @@ func TestParser_ParseUnverified(t *testing.T) {
...
+
+	_, err := jwt.Parse(tokenString, func(token *jwt.Token) (interface{}, error) {
+		return nil, nil
+	})
+
+	if err == nil {
+		t.Fatal("expected error")
+	}
+	if err.Error() != "malformed token payload: invalid character 'x' looking for beginning of value" {
+		t.Errorf("unexpected error \"%s\"", err.Error())
```
mfridman, line 301: nit. can use %q to avoid escaping double quotes.

parser.go
ripienaar: The golang slack would seem like a obvious choice - many thousands of go users there and I believe they allow project channels.

PR #86, opened 2021-08-05
When exp indicates the present, make it invalid.

soranoba:
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4

>  The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.

it is written on the reference.
So, it seems invalid when exp indicates the present. Isn't it right?

### DIFF
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -2,6 +2,7 @@ package jwt
 
 import (
 	"testing"
+	"time"
 )
 
 func TestVerifyAud(t *testing.T) {
@@ -96,6 +97,29 @@ func TestMapclaimsVerifyExpiresAtInvalidTypeString(t *testing.T) {
 	want := false
 	got := mapClaims.VerifyExpiresAt(0, false)
 
+	if want != got {
+		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
+	}
+}
+
+func TestMapclaimsVerifyExpiresAtExpire(t *testing.T) {
+	exp := time.Now().Unix()
+	mapClaims := MapClaims{
+		"exp": float64(exp),
+	}
+	want := false
+	got := mapClaims.VerifyExpiresAt(exp, true)
+	if want != got {
+		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
+	}
+
+	got = mapClaims.VerifyExpiresAt(exp + 1, true)
+	if want != got {
+		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
+	}
+
+	want = true
+	got = mapClaims.VerifyExpiresAt(exp - 1, true)
 	if want != got {
 		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
 	}
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -99,7 +99,7 @@ func TestMapclaimsVerifyExpiresAtInvalidTypeString(t *testing.T) {
 	}
 }
 
-func TestMapclaimsVerifyExpiresAtExpire(t *testing.T) {
+func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
 	exp := time.Now().Unix()
 	mapClaims := MapClaims{
 		"exp": float64(exp),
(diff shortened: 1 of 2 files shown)


lggomez:
That's a braking change on the current claims structure but there's an ongoing PR adding an API compatible implementation of a RFC7519 compatible claim at #15

cc @oxisto

lggomez:
@soranoba the PR was just merged into main, so you should be able to test this via `go get github.com/golang-jwt/jwt/v4@main`, there is a new claims type `RegisteredClaims` which performs the following RFC compliant validation: https://github.com/golang-jwt/jwt/blob/80625fb516604a88dd525c2354bd6f7d652ea165/claims.go#L237-L242

oxisto:
> @soranoba the PR was just merged into main, so you should be able to test this via `go get github.com/golang-jwt/jwt/v4@main`, there is a new claims type `RegisteredClaims` which performs the following RFC compliant validation:
> 
> https://github.com/golang-jwt/jwt/blob/80625fb516604a88dd525c2354bd6f7d652ea165/claims.go#L237-L242

I fear the new claim is just replicating the old claim behaviour. However, since we did not do a release yet, we are somewhat free to change it. I need to double check the exact wording on the RFC again and will have a look at how other libraries handle this.

soranoba:
Thank you for the information. I will check it.

soranoba:
Now that the main branch uses time.Time (It is same `dgrijalva/jwt-go` `v4.0.0-preview1`), there are no real problems with RFC violations.

In the current v4, it checks expires in 1s units, so there was a problem that it is not regarded as expire even if several ms have passed in the test code.
I wanted this case to go away, so my worries have been resolved at the main branch.

However, I think the main branch is RFC incompliant either.
```diff
- return now.Before(*exp) || now.Equal(*exp) 
+ return now.Before(*exp)
```
I think it's correct to do this, but I'm not firmly reading the RFC, so I'll leave it to you.
Thanks.

dunglas:
Your proposed patch looks correct to me @soranoba.

mfridman:
There is a merge conflict on the PR. Should we fix it up?

mfridman:
I also agree with this proposal. Likewise the time-based verification should just be:

```diff
func verifyExp(exp *time.Time, now time.Time, required bool) bool {
	if exp == nil {
		return !required
	}
- 	return now.Before(*exp) || now.Equal(*exp)
+	return now.Before(*exp)
}
```

https://github.com/golang-jwt/jwt/blob/205b3dc4bbaa1152602c7cfbfb7e18d8c357ef5d/claims.go#L241

oxisto:
> I also agree with this proposal. Likewise the time-based verification should just be:
> 
> ```diff
> func verifyExp(exp *time.Time, now time.Time, required bool) bool {
> 	if exp == nil {
> 		return !required
> 	}
> - 	return now.Before(*exp) || now.Equal(*exp)
> +	return now.Before(*exp)
> }
> ```
> 
> https://github.com/golang-jwt/jwt/blob/205b3dc4bbaa1152602c7cfbfb7e18d8c357ef5d/claims.go#L241

Yes. Looks like this is correct. I fixed the merge conflict.

map_claims_test.go
```diff
@@ -97,3 +98,26 @@ func TestMapclaimsVerifyExpiresAtInvalidTypeString(t *testing.T) {
 		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
 	}
 }
+
+func TestMapclaimsVerifyExpiresAtExpire(t *testing.T) {
```
oxisto, line 102:
```suggestion
func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
```

oxisto: Minor spelling nit

oxisto (approved):

map_claims_test.go
```diff
@@ -97,3 +98,26 @@ func TestMapclaimsVerifyExpiresAtInvalidTypeString(t *testing.T) {
 		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, got)
 	}
 }
+
+func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
```
mfridman, line 102: lgtm.

note to self to consider rewriting these tests as table tests

mfridman (approved):

PR #96, opened 2021-08-26
Add `GetSigningMethods` function

### LINKED ISSUE #63: Add API to list all available signing methods
Migrated from https://github.com/dgrijalva/jwt-go/issues/410:

>  trygvis commented on Jun 17, 2020
> 
> I'm building a CLI to create tokens, and it would be nice if there was a way to find all registered signing methods.
> 
> Something like
> 

`func GetSigningMethods() ([]SigningMethod)`

wisdommatt:
Signed-off-by: Wisdom Matthew <talk2wisdommatt@gmail.com>

resolves #63

### DIFF
--- a/signing_method.go
+++ b/signing_method.go
@@ -33,3 +33,14 @@ func GetSigningMethod(alg string) (method SigningMethod) {
 	}
 	return
 }
+
+// GetSigningMethods retrieves all registered signing methods
+func GetSigningMethods() (methods []SigningMethod) {
+	signingMethodLock.RLock()
+	defer signingMethodLock.RUnlock()
+
+	for _, method := range signingMethods {
+		methods = append(methods, method())
+	}
+	return methods
+}


mfridman:
Is there a good use-case for this?

There is a hard-coded list here: https://github.com/golang-jwt/jwt#signing-methods-and-key-types

lggomez:
> Is there a good use-case for this?
> 
> There is a hard-coded list here: https://github.com/golang-jwt/jwt#signing-methods-and-key-types

Well, from a consumers point of view it would be a nice feature to programmatically know the supported features. Case in point from the original issue I ported #63 from:

> I'm building a CLI to create tokens, and it would be nice if there was a way to find all registered signing methods.

I'm not happy with the performance of the simple implementation, although making this more performance-friendly will add more unwarranted complexity to the implementation (for instance, the consumer of this package should call this method only once and cache the returned slice of supported methods for further usage)

AlexanderYastrebov:
To address the `I'm building a CLI to create tokens, and it would be nice if there was a way to find all registered signing methods.` usecase as well as `I'm not happy with the performance of the simple implementation` I think the alternative could be to provide a list of registered `alg` names instead. 

This way the list could be used for the CLI help but also could be iterated to obtain respective `SigningMethod` if needed.

If that sounds reasonable, I would like to propose the PR that adds
```go
// GetAlgorithms returns a list of registered "alg" names
func GetAlgorithms() (algs []string) {
	signingMethodLock.RLock()
	defer signingMethodLock.RUnlock()

	for alg, _ := range signingMethods {
		algs = append(algs, alg)
	}
	return
}
```

oxisto:
Superseded by #123

PR #97, opened 2021-08-29
Harmonising capitalisation of "token" in error strings

YoogoC:


claims.go
```diff
@@ -61,7 +61,7 @@ func (c RegisteredClaims) Valid() error {
 	}
 
 	if !c.VerifyIssuedAt(now, false) {
-		vErr.Inner = fmt.Errorf("Token used before issued")
+		vErr.Inner = fmt.Errorf("token used before issued")
```
mfridman, line 64: Curious, why this PR? Are you relying on the capitilization?

I'm indifferent about this one way or the other, but I'd prefer to avoid these kinds of PRs. Closing.

Also. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

oxisto: Didn't he try to *remove* the capitalisation? It seems that we missed this one, the other `Errorf` calls are also lowercased.

mfridman: Oops, diff'd it wrong. You're right, apologies @YoogoC . Thanks for your contribution. My mistake.

mfridman (approved):

oxisto (approved):

Issue #73, opened 2021-08-03
Add support for JWK (RFC7517)


lggomez:
Migrated from https://github.com/dgrijalva/jwt-go/issues/249:

>  dgrijalva commented on Mar 8, 2018
> 
> How far down the rabbit hole should we go with this?

See the issue thread for more details

lggomez:
This is a very large undertaking in the scope of this package IMO (as it has way more responsibilities including performing HTTP requests). Should we decide to implement this, I believe it should be done in a separate package using this one

lwj5:
Consider putting it in `auth0/go-jwt-middleware`, it is a vendor-agnostic middleware. 

There's standard validation for jwt-go contributed by one of the members here but without JWK(S) support
https://github.com/auth0/go-jwt-middleware/tree/v2/validate/jwt-go

There's standard validation for jose with JWK(S) support
https://github.com/auth0/go-jwt-middleware/blob/v2/validate/josev2/josev2.go

MicahParks:
I wrote a package for creating a [`jwt.Keyfunc`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Keyfunc). It's [`github.com/MicahParks/keyfunc`](https://github.com/MicahParks/keyfunc).

This package does _not_ create a JSON Web Key set (JWKs). It's only used for validating JWTs signed by a JSON Web Keys in the set. It can optionally grab and automatically refresh the JWKs via HTTP with a couple configuration options.

It's currently a separate package that supports a few `github.com/dgrijalva/jwt-go` forks including this one. Each fork must be explicitly supported as the signature of [`jwt.Keyfunc`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Keyfunc) contains an argument that is a Go `struct`, [`*jwt.Token`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Token), not an `interface` implementation.
```go
type Keyfunc func(*Token) (interface{}, error)
```

I'd be happy to contribute a modification of my `keyfunc` package to this repository, if requested. This would not bring the repository to fully supporting JWKs though, as I don't believe it has the ability to create one.

greatcat-taihe:
It's useful parse apple sign in token.

AlexanderYastrebov:
It would be interesting to support JWKs url as a key in the `jwt` tool.

oxisto:
Closing this, as there is already a good community package for this

PR #105, opened 2021-09-17
Add jwt.Keyfunc utilities and partial JWK support

MicahParks:
# Purpose
The purpose of this Pull Request is to add [`jwt.Keyfunc`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Keyfunc) utilities found in [`github.com/MicahParks/keyfunc`](https://github.com/MicahParks/keyfunc).

These utilities include creating a [`jwt.Keyfunc`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Keyfunc) from a given set of cryptographic keys and or a local/remote JWKs. When using a remote JWKs there are additional features such as a background refresh goroutine. For additional details on the provided features, the `README.md` in the `keyfunc` directory is a good place to start.

This would add partial JWK support for this project and help make progress towards this issue: https://github.com/golang-jwt/jwt/issues/73.

It would be ideal for those who found `github.com/MicahParks/keyfunc` useful to only need to import one JWT pacakge. Here is an example of a PR that avoided importing `github.com/MicahParks/keyfunc` and resorted to copying due to the supported forks being dependencies: https://github.com/gofiber/jwt/pull/57.
> I believe I could get away with smaller changes if I would reference https://github.com/MicahParks/keyfunc directly. But that would bring additional unused legacy dependencies to this middleware and to the rest of the projects which will be using it, so I decided to copy files instead.

Additionally, if the `github.com/golang-jwt/jwt/v4` does another major semantic version update, the `github.com/MicahParks/keyfunc` will need to include support for both `/v4` and `/v5` due to the Go type system. This can be avoided if the `github.com/MicahParks/keyfunc` is included in its dependent project.

This PR only includes effectual changes in the `keyfunc` directory and does not modify anything but a comment in the original project.

# Terminology
* "JWKs" is a JSON Web Key Set (JWK Set). 
* "given keys" are cryptographic keys already in the correct Go type provided by the package user before the creation of a `keyfunc.JWKs`.

# Limitations
This package does not yet support EdDSA keys. This means if a JWKs contains a key with an `alg` of `EdDSA` will return an error during [`jwt.Parse`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Parse) if a JWT with a `kid` associated with the `alg` of `EdDSA`. I believe implementing this would be very doable, but haven't come across a JWKs that uses EdDSA keys yet. I may implement this before this PR is unmarked as a draft. The relevant RFC is here: https://datatracker.ietf.org/doc/html/rfc8037#appendix-A.4

This package does not support creating a JWKs. It only consumes a JWKs for the purpose of parsing and validating JWTs.

# Tests
Test coverage for this new `github.com/golang-jwt/jwt/v4/keyfunc` package is `>90%`. Some tests make keys to sign tokens, then parse and validate them. Other tests use hard coded tokens and JWKs.

# Examples
This PR includes an `examples` directory. Many of the examples reference this service: https://jwks-service.appspot.com/. As it states on the site:
> This service is intended to help you as you test basic JWT and JWKS interoperability.


MicahParks:
This would also provide examples that this issue requires: https://github.com/golang-jwt/jwt/issues/58

keyfunc/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+The purpose of this package is to provide a
+[`jwt.Keyfunc`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Keyfunc). This function is for parsing and validating
+JWTs.
+
+This package can consume a JSON Web Key Set (JWKs) and or a given set of key ID, `kid` to cryptographic key mappings.
+This package can additionally use a given mapping of key IDs, `kid`, to cryptographic keys parse and validate JWTs.
+
+It's common for an identity provider, such as [Keycloak](https://www.keycloak.org/)
+or [Amazon Cognito (AWS)](https://aws.amazon.com/cognito/) to expose a JWKs via an HTTPS endpoint. It is important that
```
meagar, line 11: I would use the full name before introducing the acronym, ie `JSON Web Key Set` before `JWKS`

mfridman:
Thank you for submitting a PR and writing up a detailed description. Very helpful.

I'm a bit skeptical, however, that supporting JWKs is within the scope of this `jwt` package. It seems users can already use https://github.com/MicahParks/keyfunc ?

Also, applications may choose to implement their own mechanism for fetching/storing keys which is orthogonal to verifying and parsing JWTs. The code also appears to be more complicated than it needs to be, there is a nested lock and that's a bit of a code smell to me.

But putting aside the implementation, accepting a PR like this would mean having to support this code which is another major consideration.

MicahParks:
I believe having this code in this package would benefit the pacakge users. Parsing and validating JWTs with keys found in a JWK Set is a common use case. Moreover, the documentation around creating a `jwt.KeyFunc` is sparse and the function signature uses an empty interface, `interface{}`, which has caused some confusion (GitHub issues, StackOverflow questions). The addition of "given keys" `keyfunc.NewGiven`, helps by statically typing cryptographic keys used for parsing and validating JWTs (the most common use case I see being HMAC). 

I'd be happy to discuss the benefits and drawbacks to any of the specifics of the implementation. I would also like to offer to help maintain this portion of the code, if the PR is accepted.

keyfunc/get.go
```diff
@@ -0,0 +1,223 @@
...
+	// Create some rate limiting assets.
+	var lastRefresh time.Time
+	var queueOnce sync.Once
+	var refreshMux sync.Mutex
+	if j.refreshRateLimit != nil {
+		lastRefresh = time.Now().Add(-*j.refreshRateLimit)
+	}
+
+	// Create a channel that will never send anything unless there is a refresh interval.
+	refreshInterval := make(<-chan time.Time)
```
lggomez, line 74: could this work with a time.Ticker?

lggomez:
I kinda agree with @mfridman in that the JWKS concern is in a separate scope of this package, altough related and the implementation can be decoupled in a fashion such that that the consumer decides the http Doer implementation to use (just to name one thing).

I think that, should this go forward, could go into a separate package like `golang-jwt/jwks` to ease maintenance, versioning and separation of concerns

keyfunc/get.go
MicahParks: If specifying the `RefreshInterval` of the `keyfunc.Options` had a non-zero default, yes.

Since it can be omitted with the expectation of no refresh interval, a `time.Ticker` cannot be created because it requires a `time.Duration` greater than zero.

If it were turned into a `time.Ticker` now, then the below `select` statement case would need to be guarded by an `if refreshInterval != nil` statement, which would require two `select` statements, once for each case.

MicahParks:
Seems this PR has gotten a bit stale. I'm going to close it, but please feel free to reach out if it fits inside the scope of this project and I'd be happy to re-open it.

I plan on making a handful of changes to `github.com/MicahParks/keyfunc` then releasing `v1.0.0` within a week or so.

PR #106, opened 2021-09-18
Add unit test cases to parse JWT tokens signed with ECDSA keys

sebastien-rosset:
In parse_test.go:
1. add test case where the JWT token is signed with a ECDSA private key and "EC256" is in the list of accepted algorithms.
1. add test case where the JWT token is signed with a ECDSA private key but "EC256" is NOT in the list of accepted algorithms.
1. Add validation for Token.Valid and Token.Signature in the unit tests.

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -1,6 +1,7 @@
 package jwt_test
 
 import (
+	"crypto"
 	"crypto/rsa"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


lggomez (approved):
LGTM, nice to have in the test coverage

oxisto (approved):
LGTM

PR #107, opened 2021-09-19
Improve code comments, including security consideration

sebastien-rosset:
The README file includes a security notice about validating the 'alg' claim. This PR adds code comments for the `Parse()` function to help present the information to users of the library.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -13,9 +13,9 @@ type Parser struct {
 	SkipClaimsValidation bool     // Skip claims validation during token parsing
 }
 
-// Parse parses, validates, and returns a token.
+// Parse parses, validates, verifies the signature and returns the parsed token.
 // keyFunc will receive the parsed token and should return the key for validating.
-// If everything is kosher, err will be nil
+// If everything is kosher, err will be nil.
 func (p *Parser) Parse(tokenString string, keyFunc Keyfunc) (*Token, error) {
 	return p.ParseWithClaims(tokenString, MapClaims{}, keyFunc)
 }
--- a/token.go
+++ b/token.go
@@ -29,11 +29,12 @@ type Token struct {
 	Valid     bool                   // Is the token valid?  Populated when you Parse/Verify a token
 }
 
-// New creates a new Token.  Takes a signing method
+// New creates a new Token with the specified a signing method and an empty map of claims.
 func New(method SigningMethod) *Token {
 	return NewWithClaims(method, MapClaims{})
 }
 
+// NewWithClaims creates a new Token with the specified signing method and claims.
 func NewWithClaims(method SigningMethod, claims Claims) *Token {
 	return &Token{
 		Header: map[string]interface{}{
@@ -45,7 +46,8 @@ func NewWithClaims(method SigningMethod, claims Claims) *Token {
 	}
 }
 
-// SignedString retrieves the complete, signed token
+// SignedString creates and returns a complete, signed JWT token.
+// The token is signed using the SigningMethod specified in the token.
 func (t *Token) SignedString(key interface{}) (string, error) {
 	var sig, sstr string
 	var err error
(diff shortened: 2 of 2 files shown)


vuonghx (approved):

token.go
```diff
@@ -82,9 +84,11 @@ func (t *Token) SigningString() (string, error) {
...
 }
 
-// Parse parses, validates, and returns a token.
-// keyFunc will receive the parsed token and should return the key for validating.
-// If everything is kosher, err will be nil
+// Parse parses, validates, verifies the signature and returns the parsed token.
+// keyFunc will receive the parsed token and should return the cryptographic key
+// for verifying the signature.
+// keyFunc should validate the 'alg' claim in the token matches the expected algorithm.
+// If everything is kosher, err will be nil.
```
lggomez, line 91: I would link the following for more information on the main reason, as it contains the vulnerability it may cause on unprotected code https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

Also a nitpick: I'd change _kosher_ for OK; while we both know the meaning of the word not all audiences may be familiar with it and we want the docs to be as neutral as possible on terminologies

parser.go
```diff
@@ -13,9 +13,9 @@ type Parser struct {
 	SkipClaimsValidation bool     // Skip claims validation during token parsing
 }
 
-// Parse parses, validates, and returns a token.
+// Parse parses, validates, verifies the signature and returns the parsed token.
 // keyFunc will receive the parsed token and should return the key for validating.
-// If everything is kosher, err will be nil
+// If everything is kosher, err will be nil.
```
lggomez, line 18: Nitpick: see the other comment below

lggomez (changes requested):

token.go
sebastien-rosset: Agree about the word kosher, I will change it.

sebastien-rosset: How about just removing the sentence "If everything is kosher, err will be nil.". Looking at the golang source code, functions that may return errors do not explicitly state something like "return error if xyz problem occurred". It would be fairly obvious and repetitive.

token.go
```diff
@@ -29,11 +29,12 @@ type Token struct {
 	Valid     bool                   // Is the token valid?  Populated when you Parse/Verify a token
 }
 
-// New creates a new Token.  Takes a signing method
+// New creates a new Token with the specified a signing method and an empty map of claims.
```
oxisto, line 32:
```suggestion
// New creates a new Token with the specified signing method and an empty map of claims.
```

token.go
```diff
@@ -45,7 +46,8 @@ func NewWithClaims(method SigningMethod, claims Claims) *Token {
 	}
 }
 
-// SignedString retrieves the complete, signed token
+// SignedString creates and returns a complete, signed JWT token.
```
oxisto, line 49:
```suggestion
// SignedString creates and returns a complete, signed JWT.
```

JWT token is redundant ;)

token.go
```diff
@@ -82,9 +84,12 @@ func (t *Token) SigningString() (string, error) {
 	return strings.Join(parts, "."), nil
 }
 
-// Parse parses, validates, and returns a token.
-// keyFunc will receive the parsed token and should return the key for validating.
-// If everything is kosher, err will be nil
+// Parse parses, validates, verifies the signature and returns the parsed token.
+// keyFunc will receive the parsed token and should return the cryptographic key
+// for verifying the signature.
+// keyFunc should validate the 'alg' claim in the token matches the expected algorithm.
```
oxisto, line 90: Uhm to be honest. I think this is not the correct place for this documentation or the correct way to use `keyFunc`.

The general idea is to supply a slice of `ValidMethods` to the `Parser` object, so that the matching algorithms can be checked before key func is ever reached. The idea of `keyFunc` is purely to provide a key. Sure it can be used for further custom validation, but that is not the primary point of it.

https://github.com/golang-jwt/jwt/blob/2ebb50f957d606de5909fcf9ed49f9af3bc35e97/parser.go#L30

oxisto (changes requested):

token.go
sebastien-rosset: That's a good point. IMO, the `Parse` functions defined here https://github.com/golang-jwt/jwt/blob/main/token.go#L88-L94 are partly the problem. They compel the application developer to create a token with a default parser that has a nil `ValidMethods`.
I was wondering if these functions should be marked deprecated. I.e. encourage the developer to create a parser with `ValidMethods`.

oxisto: I am working on a branch that features a similar syntax as the original `v4` branch, introducing parser options for an easier way to configure a parser. I just have to see whether we can do that completely backwards compatible or if this will be the beginning of the v5 branch.

PR #108, opened 2021-09-26
Introducing functional-style options for the Parser type

oxisto:


oxisto:
I was also thinking about getting rid of the `ParseWithClaims` functions all together and moving this to a parser option as well, but I think that may be a larger undertaking.

mfridman:
If there is a configurable parser with functional options (via variadic arguments) this opens the opportunity to add behaviour in a backwards compatible way in the current branch.

(I know I sound like a broken record w.r.t backwards compatibility, but that is the one thing above all that I care about in this package)

Do you think we should build a breaking but "nicer" API in a `/v5` release, or do we want to continue with `/v4` and add non-breaking functionality to it?

The `ParseWithClaims` was quite handy.

oxisto:
> Do you think we should build a breaking but "nicer" API in a `/v5` release, or do we want to continue with `/v4` and add non-breaking functionality to it?

I think we can sort of have both. In my opinion, the approach implemented now is completely backwards compatible and it offers a quite nice way of configuring the options

> The `ParseWithClaims` was quite handy.

I will keep it (for now) and I also added variadic arguments to it, so it can be fed with the same `WithXXX` functional options as the pure `Parse` function.

PR #115, opened 2021-10-22
no need for string slice and call to strings.join

moneszarrugh:
this cleans up the code a bit and avoids unnecessary memory allocations

### DIFF
--- a/token.go
+++ b/token.go
@@ -63,35 +63,34 @@ func NewWithClaims(method SigningMethod, claims Claims) *Token {
 // SignedString creates and returns a complete, signed JWT. The token is signed
 // using the SigningMethod specified in the token.
 func (t *Token) SignedString(key interface{}) (string, error) {
-	var sig, sstr string
-	var err error
-	if sstr, err = t.SigningString(); err != nil {
+	sstr, err := t.SigningString()
+	if err != nil {
 		return "", err
 	}
-	if sig, err = t.Method.Sign(sstr, key); err != nil {
+
+	sig, err := t.Method.Sign(sstr, key)
+	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


adwpc (dismissed):

AlexanderYastrebov:
> and avoids unnecessary memory allocations

Would it be possible to add a benchmark to demonstrate that change indeed reduces allocations?

lggomez:
@moneszarrugh @AlexanderYastrebov There is an example on #34 (which is stale but I came back to take a look at and should be merged in the meantime before this one goes to the same code)

moneszarrugh:
> @moneszarrugh @AlexanderYastrebov There is an example on #34 (which is stale but I came back to take a look at and should be merged in the meantime before this one goes to the same code)

Sorry for keeping this stale, as I have no freetime to benchmark and test it.
If anyone is interested in adding these feel free to do it.

oxisto:
> > @moneszarrugh @AlexanderYastrebov There is an example on #34 (which is stale but I came back to take a look at and should be merged in the meantime before this one goes to the same code)
> 
> Sorry for keeping this stale, as I have no freetime to benchmark and test it. If anyone is interested in adding these feel free to do it.

So I did some checking and in fact it does not have less allocations, however, the code is cleaner and easier to understand and thus it makes sense to accept it.

Before:
```
BenchmarkHS512Signing-8   	 1604331	       715.2 ns/op	    1777 B/op	      26 allocs/op
```

After:
```
BenchmarkHS512Signing-8   	 1616665	       702.9 ns/op	    1777 B/op	      26 allocs/op
```

oxisto (approved):

mfridman (approved):

Issue #119, opened 2021-11-04
Code comment for jwt.ParseRSAPublicKeyFromPEM incorrect


twocs:
The code comment for rsa_utils.go says it's for a PKCS1 or PKCS8 public key (`ParseRSAPublicKeyFromPEM parses a PEM encoded PKCS1 or PKCS8 public key
ParseRSAPublicKeyFromPEM parses a PEM encoded PKCS1 or PKCS8 public key`), but the function uses x509.ParsePKIXPublicKey instead. I was unable to get a PKCS1 public key despite the code comment, and basically just copied the entire function and swapped it from `x509.ParsePKIXPublicKey` to `x509.ParsePKCS1PublicKey`, which works. I am not clear but it seems that it's not matching the code comment and it took me quite some time to figure out why I couldn't parse my RSA Public Key (it wasn't accepted on jwt.io).

I suspect that the problem is not with the code comment, but with the implementation. Or am I doing something wrong that the parsing of a pem with PKCS1 encoding doesn't work using  `x509.ParsePKIXPublicKey` but does work with `x509.ParsePKCS1PublicKey`?

https://github.com/golang-jwt/jwt/blob/2ebb50f957d606de5909fcf9ed49f9af3bc35e97/rsa_utils.go#L79
```golang
// ParseRSAPublicKeyFromPEM parses a PEM encoded PKCS1 or PKCS8 public key
func ParseRSAPublicKeyFromPEM(key []byte) (*rsa.PublicKey, error) {
	var err error

	// Parse PEM block
	var block *pem.Block
	if block, _ = pem.Decode(key); block == nil {
		return nil, ErrKeyMustBePEMEncoded
	}

	// Parse the key
	var parsedKey interface{}
	if parsedKey, err = x509.ParsePKIXPublicKey(block.Bytes); err != nil {
		if cert, err := x509.ParseCertificate(block.Bytes); err == nil {
			parsedKey = cert.PublicKey
		} else {
			return nil, err
		}
	}
```

If we compare to https://github.com/golang-jwt/jwt/blob/2ebb50f957d606de5909fcf9ed49f9af3bc35e97/rsa_utils.go#L17
private key explicitly tries both functions.
```
	if parsedKey, err = x509.ParsePKCS1PrivateKey(block.Bytes); err != nil {
		if parsedKey, err = x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
			return nil, err
		}
	}
```

oxisto:
[...]

PR #120, opened 2021-11-06
enable jwt.ParsePublicKeyFromPEM to parse PKCS1 Public Key

twocs:
#119 
The code comment for jwt.ParsePublicKeyFromPEM stated that it could parse PKCS1 public keys.
However, it couldn't. I created a unit test that generates a public PKCS1 key and the function cannot pass the test. 
I have added an extra attempt to parse the public key via the x509.ParsePKCS1PublicKey function, but also retained previous functionality that attempts to parse it from PKIX or certificate first.


rsa_test.go
```diff
@@ -155,6 +160,25 @@ func TestRSAKeyParsing(t *testing.T) {
 
 }
 
+func TestRSAParsePublicKeyFromPEM_PKCS1(t *testing.T) {
```
oxisto, line 163: Could we integrate this into the existing `TestRSAKeyParsing` function? We could then rework the existing test functions to use the table-driven test paradigm.

rsa_utils.go
```diff
@@ -91,7 +92,9 @@ func ParseRSAPublicKeyFromPEM(key []byte) (*rsa.PublicKey, error) {
 		if cert, err := x509.ParseCertificate(block.Bytes); err == nil {
 			parsedKey = cert.PublicKey
 		} else {
-			return nil, err
+			if parsedKey, err = x509.ParsePKCS1PublicKey(block.Bytes); err != nil {
+				return nil, fmt.Errorf("%v - %w", ErrNotRSAPublicKey.Error(), err)
```
oxisto, line 96: this sort of breaks our existing error pattern. this should probably just return `ErrKeyMustBePEMEncoded`.

oxisto (approved):

PR #121, opened 2021-11-06
Allow `none` algorithm in jwt command

AlexanderYastrebov:
Usage:
```
$ echo '{"foo":"bar"}' | ./jwt -alg none -sign - | ./jwt -alg none -verify -
```

I know that `none` method should not be used in real applications but I think `jwt` as a tool should support it.

Signed-off-by: Alexander Yastrebov <yastrebov.alex@gmail.com>

### DIFF
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -128,6 +128,9 @@ func verifyToken() error {
 
 	// Parse the token.  Load the key from command line option
 	token, err := jwt.Parse(string(tokData), func(t *jwt.Token) (interface{}, error) {
+		if isNone() {
+			return jwt.UnsafeAllowNoneSignatureType, nil
+		}
 		data, err := loadData(*flagKey)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
@oxisto `jwt` is a versatile tool that could be used to show, sign and verify tokens. It could be used outside of go ecosystem, e.g. by security researchers. I think `node` algorithm should be supported for completeness. I have also updated PR to not require `-key` flag.

oxisto:
> @oxisto `jwt` is a versatile tool that could be used to show, sign and verify tokens. It could be used outside of the go ecosystem, e.g. by security researchers. I think `none` algorithm should be supported for completeness. I have also updated PR to not require `-key` flag.

Good points. Fair enough :)

oxisto:
Could we print out some warning on stderr if none is used? Or would that break the functionality of the tool?

AlexanderYastrebov:
> Could we print out some warning on stderr

I can add the warning but I do not think it is necessary. `none` is not the default - user has to explicitly choose it. The tool also does not print warnings e.g. if HS256 key is empty or short.

oxisto (approved):

mfridman (approved):

Issue #16, opened 2021-05-28
Bring over ValidationHelper functionality from v4 branch


oxisto:
The v4 branch has a nice validation helper struct which combines a lot of the functionality of the `Validate` functions. It might be nice to re-implement that and slowly move existing functionality to it.

This would include things that have been on the wishlist in other forks, such as having leeway when validating timestamps (see https://github.com/form3tech-oss/jwt-go/pull/12).

mcquackers:
I was wondering about the implementation of the leeway:
Why does the incorporation of leeway depend on the validating server instead of the authorizing client?
To me, it seems like it would be cleaner for the JWT to be the source of truth, with leeway built into the iat/exp

johannesvietze:
Also here because of #98, which just caused a downtime for over an hour on our service. Thanks for working on this already!

miparnisari:
> Currently I am waiting for #108 to be merged. 

it's been merged 👀  :)

oxisto:
I had a quick look yesterday and I am not sure if we can port this over in a backwards compatible manner, since the interface of `Claims` needs to be changed. Especially the `Valid` function needs to have an additional parameter for the `ValidationHelper`. I guess I will first focus on the functionality and then see whether we can port this to v4 or if this will then introduce a v5 branch with breaking changes.

ksegun:
This is something we relied on from the old library as we ran on the preview version and would help us move to adopt this port.

arinto:
Yes, we also rely in this functionality from the old library :)

@oxisto , regarding your statement below: 
>  I guess I will first focus on the functionality and then see whether we can port this to v4 or if this will then introduce a v5 branch with breaking changes.

So, what are the intended functionality that are targeted in v4.x?

oxisto:
Please see https://github.com/golang-jwt/jwt/discussions/211 for an ongoing discussion how to proceed.

oxisto:
Fixed by #234

PR #139, opened 2021-12-06
feat: port clockskew support

ksegun:
The approach is similar to that used by the golang mongodb driver to maintain backwards compatibility

This should mitigate https://github.com/golang-jwt/jwt/issues/16


oxisto:
> The approach is similar to that used by the golang mongodb driver to maintain backwards compatibility
> 
> This should mitigate #16

Thanks for the contribution, very much welcome, this is a feature that was on the wish list for a long time. This is definitely something I also wanted to explore as part of the new validator helper approach. I promise that I will find some time on the upcoming weekend to have a look at this, unfortunately we are quite in a end-of-the-year rush here. I also want to align this with the work I have previously started as well.

mfridman:
Agree with much of what @oxisto said. (I'm also a bit swamped)

My only 2 cents is we should take our time and think through whether this is the correct API and ensure this is a backwards compatible change. (not saying it's not, just that we think through this).

We should also use functional options instead of a variadic pointer to a struct.

ksegun:
Agree, we would prefer to have functional options for those of us stuck on the v4 preview in the old library we are quite happy with them. 

As we evaluated our options on whether to port to a different library with skew support or this, this was a trade-off that we thought could be adequate until functional options are made available and there could be far easier ways as well now that I consider it e.g. an environment variable.

We fully support whatever approach is decided on.

claims.go
```diff
@@ -9,7 +9,7 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
-	Valid() error
+	Valid(options ...*ClaimsValidationOptions) error
```
oxisto, line 12: As @mfridman already mentioned, would it be possible to turn these into functional style options? Similar to what we have in https://github.com/golang-jwt/jwt/blob/main/parser_option.go

claims_option.go
```diff
@@ -0,0 +1,31 @@
+package jwt
+
+import "time"
+
+// ClaimsValidationOptions represents options that can be used for claims validation
+type ClaimsValidationOptions struct {
```
oxisto, line 6: I would call this `ValidatorOptions`, similar to the `ParserOptions` we have

oxisto (changes requested):

claims_option.go
ksegun: Renamed

claims.go
ksegun: The `WithLeeway` function is functional style https://github.com/golang-jwt/jwt/pull/139/files#diff-72f644d282bce7e136d0f762de1b9d5f3e447bfa88e5e0c4974ac2479107ddc9R33-R37 

It is far easier to set this as an attribute of the parser which is what the old library did, but it is not a backward compatible change and given how the `Valid` function is called (https://github.com/golang-jwt/jwt/pull/139/files#diff-83eb8e32639d01cf443d6d8bde24c1c8be78766090d8c5f8586c36250cfedca6R87)  I am frankly unsure how you make the claims validation functional style, however, I would be happy to make the change with your input.

ksegun:
@oxisto this is now all functional style

oxisto:
> @oxisto this is now all functional style

Nice! I will have a detailed look on this in the upcoming days. Thanks for the improvements

validator_option.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "time"
+
+// ValidatorOption is used to implement functional-style options that modify the behavior of the parser. To add
+// new options, just create a function (ideally beginning with With or Without) that returns an anonymous function that
+// takes a *ValidatorOptions type as input and manipulates its configuration accordingly.
+type ValidatorOption func(*ValidatorOptions)
+
+// ValidatorOptions represents options that can be used for claims validation
+type ValidatorOptions struct {
```
oxisto, line 11: Suggestion: This type should be called `Validator` or rather `validator`. Additionally, I would make this type unexported, at least for now, until our API is more stabilised.

validator_option.go
```diff
@@ -0,0 +1,21 @@
+package jwt
+
+import "time"
+
+// ValidatorOption is used to implement functional-style options that modify the behavior of the parser. To add
+// new options, just create a function (ideally beginning with With or Without) that returns an anonymous function that
+// takes a *ValidatorOptions type as input and manipulates its configuration accordingly.
+type ValidatorOption func(*ValidatorOptions)
```
oxisto, line 8: maybe ValidationOption instead? Probably a toin coss at some point. But the way I see it, this function MUST be exported (or must it) so we should choose a name we are happy with.

cc @mfridman

parser.go
```diff
@@ -22,6 +22,8 @@ type Parser struct {
 	//
 	// Deprecated: In future releases, this field will not be exported anymore and should be set with an option to NewParser instead.
 	SkipClaimsValidation bool
+
+	options []ValidatorOption
```
oxisto, line 26: validationOptions?

claims.go
```diff
@@ -254,11 +270,12 @@ func verifyIat(iat *time.Time, now time.Time, required bool) bool {
 	return now.After(*iat) || now.Equal(*iat)
 }
 
-func verifyNbf(nbf *time.Time, now time.Time, required bool) bool {
+func verifyNbf(nbf *time.Time, now time.Time, required bool, clockSkew time.Duration) bool {
```
oxisto, line 273: `skew` would probably be more in line with the other short argument names

claims.go
```diff
@@ -240,11 +256,11 @@ func verifyAud(aud []string, cmp string, required bool) bool {
 	return result
 }
 
-func verifyExp(exp *time.Time, now time.Time, required bool) bool {
+func verifyExp(exp *time.Time, now time.Time, required bool, clockSkew time.Duration) bool {
 	if exp == nil {
 		return !required
 	}
-	return now.Before(*exp)
+	return now.Before((*exp).Add(+clockSkew))
```
oxisto, line 263: Wouldn't it be more logically to Add/Sub the clock skew from `now`? Same goes for verifyNbf

claims.go
```diff
@@ -9,7 +9,7 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
-	Valid() error
+	Valid(options ...ValidatorOption) error
```
oxisto, line 12: for the sake of brevity and consistency it should be `opts`

oxisto (commented):
Just some minor things, but I want to have at least a small discussion on naming before we merge this (as this impacts the public API).

Good job on making this non-breaking!

cc @mfridman

claims.go
ksegun: The math is sort of backward, decided to leave it alone. Unit tests confirm this is accurate.

validator_option.go
ksegun: Only needs to be exported if we want to allow usage in the call to Valid outside of the parser.  Changed so nothing is exported except the `WithLeeway` function in the parser.

oxisto: Good idea. Probably the best approach until this is stabilized from an API point of view.

claims.go
oxisto: Should we call it leeway as a parameter then? Similar to the validator option. Then it would also make sense to add it to exp? What do you think?

ksegun: Yes, we could I used skew here because that is typically how it is described in the literature and I felt it would make it clear in the code and retained WithLeeway because that is how the original library had it.

oxisto (dismissed):
LGMT. This is / should be 100 % backwards compatible and fixes a long-wanted feature. Furthermore, it allows us to go into the direction of a validation helper that fine-tunes validation of the parser, also completely backwards compatible and incrementally. All new types are unexported, so we are free to change / stabilize the API. All function signature changes are variadic.

I definitely want to get some more votes on this before we merge though

cc @golang-jwt/maintainers

validator_option.go
```diff
@@ -0,0 +1,21 @@
+package jwt
+
+import "time"
+
+// validationOption is used to implement functional-style options that modify the behavior of the parser. To add
+// new options, just create a function (ideally beginning with With or Without) that returns an anonymous function that
+// takes a *ValidatorOptions type as input and manipulates its configuration accordingly.
```
mfridman, line 7: I think this is a typo, s/`*ValidatorOptions`/`*validator`

claims.go
```diff
@@ -9,7 +9,7 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
-	Valid() error
+	Valid(opts ...validationOption) error
```
mfridman, line 12: I think we should add a comment, either on this interface (or maybe even better) around the `validationOption` explaining why it is not an exported identifier. 

Looking at this method signature I'd be a little confused.

mfridman (dismissed):
lgtm.

oxisto:
@mfridman I have tweaked the documentation. I think we can merge it then? We are still free to change the internal names in the future.

oxisto (approved):

mfridman (approved):
Sounds good to me. 

Outside the scope of this PR, we might want to revisit all these functional options in a future breaking release. We should start a discussion about that?

oxisto:
> Sounds good to me.
> 
> Outside the scope of this PR, we might want to revisit all these functional options in a future breaking release. We should start a discussion about that?

Ideally, we already started them in a way that we are happy with so not too many breaking changes are necessary. That is the main reason why I wanted to keep most of the validation structs and functions unexpected, so we can easily change them again internally.

claims.go
kolatat: Since the interface changed to include an unexported type, it is impossible to implement user-defined `Claims` (and `Token` since it contains `Claims`)?

oxisto: You should be able to implement user-defined `Claims` by embedding either `RegisteredClaims` or the old `StandardClaims`. It could be impossible to completely implement a different claims structure that does not re-use `RegisteredClaims`, but that is not intended anyway.

See our example here https://github.com/golang-jwt/jwt/blob/d489c99d3e8a82bd47dcb3e4880414f7f747eaa7/example_test.go#L33-L39

gustavosbarreto: This breaks backwards compatibility for anyone using older `v3.x.y` and `dgrijalva/jwt-go`:

```
cannot use &"github.com/golang-jwt/jwt/v4".MapClaims{} (type *"github.com/golang-jwt/jwt/v4".MapClaims) as type "github.com/golang-jwt/jwt".Claims in field value:
	*"github.com/golang-jwt/jwt/v4".MapClaims does not implement "github.com/golang-jwt/jwt".Claims (wrong type for Valid method)
		have Valid(..."github.com/golang-jwt/jwt/v4".validationOption) error
		want Valid() error
```
Can you please consider adding a new function (`ValidWithOptions`) instead of changing `Valid()` funtion signature?

PR #162, opened 2022-02-08
Set json encoding precision

### LINKED ISSUE #158: NumericDate marshals to JSON with decimals for large dates
The following code results in str being equal to "5243700879.000001"

```go
buf, _ := jwt.NewNumericDate(time.Unix(5243700879, 0)).MarshalJSON()
str := string(buf)
```

I expected the JSON representation to be without decimals.

mfridman:
Crude implementation to fix #158.

Alternatively we could switch on `TimePrecision` to set the precision, but this might be confusing and deferring to `strconv.FormatFloat(...)` with `-1` seems like the right approach.

Also, what happens if someone sets the default to minutes .. should we need to account for that?


oxisto:
> Crude implementation to fix #158.
> 
> Alternatively we could switch on `TimePrecision` to set the precision, but this might be confusing and deferring to `strconv.FormatFloat(...)` with `-1` seems like the right approach.

I am thinking whether we could calculate the float precision. Need to spin my head around that 

> 
> Also, what happens if someone sets the default to minutes .. should we need to account for that?

No, float precision 0 should be the "minimum" precision (seconds).

parser_test.go
```diff
@@ -642,3 +642,49 @@ func benchmarkSigning(b *testing.B, method jwt.SigningMethod, key interface{}) {
 		}
 	})
 }
+
+func TestNewNumericDateMarshalJSON(t *testing.T) {
```
oxisto, line 646: We should probably do that consistently at some point, but the convention seems to be `TestNumericDate_MarshalJSON`

parser_test.go
```diff
@@ -642,3 +642,49 @@ func benchmarkSigning(b *testing.B, method jwt.SigningMethod, key interface{}) {
...
+		{time.Unix(5243700879, 0), "5243700879.000001", time.Microsecond},
+		{time.Unix(5243700879, 0), "5243700879.000001", time.Nanosecond},
+		//
+		{time.Unix(4239425898, 0), "4239425898", time.Second},
+		{time.Unix(4239425898, 0), "4239425898", time.Millisecond},
+		{time.Unix(4239425898, 0), "4239425898", time.Microsecond},
+		{time.Unix(4239425898, 0), "4239425898", time.Nanosecond},
+		//
+		{time.Unix(0, 1644285000210402000), "1644285000", time.Second},
+		{time.Unix(0, 1644285000210402000), "1644285000.2099998", time.Millisecond},
```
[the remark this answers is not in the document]
oxisto: Problem is that the calculation for it is basically log_10(time.Precision / time.Second), which might not be as fast as an if-?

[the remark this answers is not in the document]
oxisto: In Go terms it would be `int(math.Log10(float64(TimePrecision) / time.Hour.Seconds()))` if you want to give it a spin.

parser_test.go
oxisto: Also, this unit test should be in `types_test.go`

oxisto:
I have created an old school diff for a calculation approach here:

```diff
diff --git a/parser_test.go b/parser_test.go
index 87a18e5..a224fc9 100644
--- a/parser_test.go
+++ b/parser_test.go
@@ -657,24 +657,24 @@ func TestNewNumericDateMarshalJSON(t *testing.T) {
 		precision time.Duration
 	}{
 		{time.Unix(5243700879, 0), "5243700879", time.Second},
-		{time.Unix(5243700879, 0), "5243700879.000001", time.Millisecond},
+		{time.Unix(5243700879, 0), "5243700879.000", time.Millisecond},
 		{time.Unix(5243700879, 0), "5243700879.000001", time.Microsecond},
-		{time.Unix(5243700879, 0), "5243700879.000001", time.Nanosecond},
+		{time.Unix(5243700879, 0), "5243700879.000000954", time.Nanosecond},
 		//
 		{time.Unix(4239425898, 0), "4239425898", time.Second},
-		{time.Unix(4239425898, 0), "4239425898", time.Millisecond},
-		{time.Unix(4239425898, 0), "4239425898", time.Microsecond},
-		{time.Unix(4239425898, 0), "4239425898", time.Nanosecond},
+		{time.Unix(4239425898, 0), "4239425898.000", time.Millisecond},
+		{time.Unix(4239425898, 0), "4239425898.000000", time.Microsecond},
+		{time.Unix(4239425898, 0), "4239425898.000000000", time.Nanosecond},
 		//
 		{time.Unix(0, 1644285000210402000), "1644285000", time.Second},
-		{time.Unix(0, 1644285000210402000), "1644285000.2099998", time.Millisecond},
+		{time.Unix(0, 1644285000210402000), "1644285000.210", time.Millisecond},
 		{time.Unix(0, 1644285000210402000), "1644285000.210402", time.Microsecond},
-		{time.Unix(0, 1644285000210402000), "1644285000.210402", time.Nanosecond},
+		{time.Unix(0, 1644285000210402000), "1644285000.210402012", time.Nanosecond},
 		//
 		{time.Unix(0, 1644285315063096000), "1644285315", time.Second},
 		{time.Unix(0, 1644285315063096000), "1644285315.063", time.Millisecond},
 		{time.Unix(0, 1644285315063096000), "1644285315.063096", time.Microsecond},
-		{time.Unix(0, 1644285315063096000), "1644285315.063096", time.Nanosecond},
+		{time.Unix(0, 1644285315063096000), "1644285315.063096046", time.Nanosecond},
 	}

 	for i, tc := range tt {
diff --git a/types.go b/types.go
index a5ef506..0e83229 100644
--- a/types.go
+++ b/types.go
@@ -49,10 +49,8 @@ func newNumericDateFromSeconds(f float64) *NumericDate {
 // MarshalJSON is an implementation of the json.RawMessage interface and serializes the UNIX epoch
 // represented in NumericDate to a byte array, using the precision specified in TimePrecision.
 func (date NumericDate) MarshalJSON() (b []byte, err error) {
-	prec := -1
-	if TimePrecision == time.Second {
-		prec = 0
-	}
+	prec := int(math.Log10(float64(time.Second) / float64(TimePrecision)))
+
 	f := float64(date.Truncate(TimePrecision).UnixNano()) / float64(time.Second)

 	return []byte(strconv.FormatFloat(f, 'f', prec, 64)), nil
```

parser_test.go
mfridman: renamed and moved in b72bb8e626a9dee1c43196840f2721af0b9ee424

parser_test.go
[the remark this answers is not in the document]
mfridman: Ye, as I was thinking this through and using various inputs it felt wrong to use `-1` when the precision is set. Exactly as you pointed out.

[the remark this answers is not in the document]
mfridman: This does raise the question, should

```
{time.Unix(5243700879, 0), "5243700879.000", time.Millisecond}
```

return "5243700879" or "5243700879.000"?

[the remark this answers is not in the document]
mfridman: Another thing to consider is whether we want this behaviour change, note after updating the tests with new implementation this failed:

https://github.com/golang-jwt/jwt/runs/5119881167?check_suite_focus=true#step:4:204

And here is the fixed test, but need to think about this one (it's getting late for me right now).

https://github.com/golang-jwt/jwt/pull/162/commits/69555c5b64bd9317109d943e68ae070490089703

types_test.go
```diff
@@ -18,12 +18,10 @@ func TestNumericDate(t *testing.T) {
 
 	jwt.TimePrecision = time.Microsecond
 
-	raw := `{"iat":1516239022,"exp":1516239022.12345}`
+	raw := `{"iat":1516239022.000000,"exp":1516239022.123450}`
```
mfridman, line 21: I suppose this makes sense. 

I personally haven't encountered a need to set the precision in any of the projects I've worked on, and second granularity (without precision) has been sufficient.

Maybe we're overthinking it, and the alternative solution is to _always_ set the default precision to 0.

oxisto: Yes it does make sense. We are setting precision to microsecond and are serialising as microsecond. Sounds good for me.

types.go
```diff
@@ -49,9 +49,10 @@ func newNumericDateFromSeconds(f float64) *NumericDate {
 // MarshalJSON is an implementation of the json.RawMessage interface and serializes the UNIX epoch
 // represented in NumericDate to a byte array, using the precision specified in TimePrecision.
 func (date NumericDate) MarshalJSON() (b []byte, err error) {
+	prec := int(math.Log10(float64(time.Second) / float64(TimePrecision)))
```
oxisto, line 52: A minor thing here: We probably need to also set a minimum precision to 0 because if you set it to TimePrecision of Seconds, log10 will return -1 which means unlimited precision. Otherwise we could return an error if you set precision above seconds (which does not really make sense anyway).

oxisto (commented):
Minor things, but I like it :)

parser_test.go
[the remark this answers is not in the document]
oxisto: See comment below. The change looks ok for me, since the original behaviour in the test was already a bit flaky.

oxisto:
@mfridman Do you remember who had brought up the original issue of float precision? I would need to look it up. Might be interesting to get some feedback from them, to see if their use case still works. My gut feeling is that this won't break anything but be potentially more precise.

mfridman:
> @mfridman Do you remember who had brought up the original issue of float precision (not the issue in the PR but the original move from int to float)? I would need to look it up. Might be interesting to get some feedback from them, to see if their use case still works. My gut feeling is that this won't break anything but be potentially more precise.

Are you thinking of [this merged PR](https://github.com/golang-jwt/jwt/pull/112).

oxisto:
> > @mfridman Do you remember who had brought up the original issue of float precision (not the issue in the PR but the original move from int to float)? I would need to look it up. Might be interesting to get some feedback from them, to see if their use case still works. My gut feeling is that this won't break anything but be potentially more precise.
> 
> Are you thinking of [this merged PR](https://github.com/golang-jwt/jwt/pull/112)?

This one: https://github.com/golang-jwt/jwt/pull/9 it was @dunglas.

oxisto (approved):

types.go
mfridman: Good catch. I hope no one is setting the precision to minutes or higher. This is something the specification could have been more explicit about.

dunglas:
@oxisto the problem I fixed was when decoding JWTs issued by other libraries such as `lcobucci/jwt`, which use floating-point numbers by default.

This change looks good to me.

dunglas (approved):

PR #163, opened 2022-02-08
cmd: add handling of HS keys

greenpau:
cmd: add handling of HS keys

Signed-off-by: Paul Greenberg <greenpau@outlook.com>


greenpau:
> IMO HS already works as expected:

@AlexanderYastrebov , it very well may be 😄 But, in imho, `<(echo -n foobar)` is not a super user-friendly way to pass keys.

PR #165, opened 2022-02-13
fix: expired token error message

ydylla:
Hi,
I noticed that the arguments in the error message for expired tokens from RegisteredClaims are in the wrong order.
Currently, it prints `13.3528857s by token is expired` but it should probably be `token is expired by 13.3528857s`.
This pull request corrects that.

Edit: Also fixed it in StandardClaims


ripienaar (approved):
LGTM, was about to come send this PR also

oxisto (approved):
Looks good.

mfridman (approved):

PR #173, opened 2022-03-01
Add go1.18 to ci pipeline

mfridman:
Checking pipeline with upcoming major Go version (1.18rc1)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [1.15, 1.16, 1.17, "1.18.0-rc1"]
+        go: [1.15, 1.16, 1.17, 1.18]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [1.15, 1.16, 1.17, "1.18.0-rc1"]
+        go: [1.15, 1.16, 1.17, 1.18]
```
mfridman, line 28: Let's drop 1.15 ?

oxisto: Should we bump the `go.mod` version to Go 1.16 then as well?

oxisto: Or maybe release a new version first, especially with #175 and #139?

mfridman:
I bumped go.mod to 1.16 minimum, and also updated the GitHub actions to v3 (both checkout and go-setup).

Note that stable was removed, so future tags can be used with semver

https://github.com/actions/setup-go/releases/tag/v3.0.0

oxisto (approved):

Issue #98, opened 2021-08-30
Token used before issued


JorritSalverda:
Having switched from `github.com/dgrijalva/jwt-go` to `github.com/golang-jwt/jwt/v4` I still get `token used before issued` errors due to clock skew, even though @dgrijalva said this check would be removed from v4, in line with jwt spec. See    https://github.com/golang-jwt/jwt/blob/3f50a786ff28f918ba8e7fb9183390b11483e755/claims.go#L63-L66

However I see that check is still executed at https://github.com/dgrijalva/jwt-go/issues/314#issuecomment-481789374. Is this dependent on https://github.com/golang-jwt/jwt/issues/16 getting resolved? Until that's resolved can the `VerifyIssuedAt` check just be dropped? Or any other suggested workarounds?

oxisto:
Yes, this is somewhat dependent on #16. Meanwhile you can use `jwt.TimeFunc` to account for clock skew.

https://github.com/golang-jwt/jwt/blob/2ebb50f957d606de5909fcf9ed49f9af3bc35e97/token.go#L10-L13

JorritSalverda:
Thanks, I'll test with

```go
jwt.TimeFunc = func() time.Time {
		return time.Now().UTC().Add(time.Second * 20)
	}
```

to see if it gets rid of these errors.

mfridman:
Hey @JorritSalverda did that solve your issue?

@oxisto Were we still interested in https://github.com/golang-jwt/jwt/issues/16

JorritSalverda:
Hi @mfridman for some reason it didn't help and i resorted to jumping through a lot of hoops to ignore this type of validation error, see https://github.com/estafette/estafette-ci-api/blob/main/pkg/clients/bitbucketapi/client.go#L277-L309

Looking at my logs the now value is actually larger than iat, so have to figure out how to get the actual value for now used during evaluation.

mfridman:
Thanks for the insight. I think introducing (configurable) leeway could help with these sorts of issues.

Side note, every-time I see folks having to resort  binary operators to determine the error it makes me cringe, because it's so prone to error.

shogo82148:
Is the error "token used before issued" really needed?

RFC7519 says that:

> https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4
> The "exp" (expiration time) claim identifies the expiration time on or after which the JWT MUST NOT be accepted for processing.

RFC7519 also says that:
[...]

PR #175, opened 2022-03-08
Make issued at (iat) claim validation optional

ksegun:
Based on https://datatracker.ietf.org/doc/html/rfc7519/#section-4.1.6 we should only validate the type for map claims.

The current behavior remains unchanged and this allows the caller to disable the check.

This PR also adds in the audience validation functional options from the V4 version of the original library.

Mitigates https://github.com/golang-jwt/jwt/issues/98


oxisto (commented):
Nice! Thank you for this PR, this has also been one of the longer outstanding issues and it seems to work quite nicely with our nice non-breaking validation options.

I will dedicate some time over the next days to look through it more carefully. One remark up front: Is it possible to make this the default (as this is what existing users of the library are expecting).

ksegun:
I sure can and you should consider  adding leeway support if it continues to be time based, effectively this is similar to the 'nbf' check.

oxisto:
> You should consider adding leeway support if it continues to be time based, effectively this is similar to the 'nbf' check.
> 
> Just to clarify, you want the default to be that this is no longer checked? If so that is what this PR does.

Sorry, I was in a rush this morning when I typed this. I meant the default to be the "old" behaviour. I would propose the following: Keeping IAT checking enabled (per default) in the v4 branch, with an extra remark that we will deprecate this behaviour in v5 and then switch the default to disabled in v5. Does that make sense?

ksegun:
Ok, would you like for me to add leeway support?

claims.go
```diff
@@ -130,6 +134,27 @@ func (c *RegisteredClaims) VerifyIssuer(cmp string, req bool) bool {
 	return verifyIss(c.Issuer, cmp, req)
 }
 
+func (c *RegisteredClaims) validateAudience(req bool, opts ...validationOption) bool {
```
oxisto, line 137: Is it somehow possible to move this function to our `validation` struct? Otherwise we are duplicating a lot of code. This would also make sense to move the other validate/verify functions there in the long run.

parser_option.go
```diff
@@ -36,3 +36,26 @@ func WithLeeway(d time.Duration) ParserOption {
 		p.validationOptions = append(p.validationOptions, withLeeway(d))
 	}
 }
+
+// WithoutIssuedAt is an option to disable the validation of the issued at (iat) claim.
+// The current `iat` time based validation is planned to be deprecated in v5
+
+func WithoutIssuedAt() ParserOption {
```
oxisto, line 43: WithoutIssuedAtValidation? Although that is a bit long

claims.go
ksegun: Unfortunately, it cannot be moved as each `VerrifyAudience` call is a receiver method on the respective struct. I refactored most of the code into a validator function to reduce duplication.

b4k3r:
@oxisto any update on it?

oxisto:
Closing this in favour of our new v5 branch (#234) which include this feature.

PR #176, opened 2022-03-09
Port audience validation functions from v4

ksegun:


oxisto:
Do you have any preference, whether we should deal with this one or #175 first? Or was there a particular reason not to combine both PRs?

ksegun:
I was simply separating concerns. Happy to combine into one PR if you prefer. I saw there was an ask for this so I went ahead an added it.

oxisto:
> I was simply separating concerns. Happy to combine into one PR if you prefer. I saw there was an ask for this so I went ahead and added it.

I think combining them into one PR makes the review easier (for me at least), otherwise things will probably be doubled. So if there is no strong need to merge the IAT one first, I would suggest to combine them.

ksegun:
Combined with #175

PR #184, opened 2022-03-18
Revert "feat: port clockskew support (#139)"

### LINKED ISSUE #183: v4.4.0 breaks backwards compatibilty
After upgrading from 4.3.0 to 4.4.0, my custom implementation of the `Claims` interface is broken:

```
	*Claims does not implement jwt.Claims (wrong type for Valid method)
		have Valid() error
		want Valid(opts ...jwt.validationOption) error
```

mfridman:
This reverts commit d489c99d3e8a82bd47dcb3e4880414f7f747eaa7.

Fix #183

Reason: this is a breaking change within a major version (v4). Users might rely on the public interface and the change in signature may cause issues.


mfridman:
If this is merged, we'll push a `v4.4.1` which contains the revert + retraction directive.

nguyengiabk (approved):

claims.go
```diff
@@ -9,10 +9,7 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
-	// Valid implements claim validation. The opts are function style options that can
-	// be used to fine-tune the validation. The type used for the options is intentionally
-	// un-exported, since its API and its naming is subject to change.
-	Valid(opts ...validationOption) error
+	Valid() error
```
gustavosbarreto, line 12: In order to maintain backwards compatibility, a `ValidWithOptions(...)` could be added to the public Interface.

gustavosbarreto (approved):

claims.go
oxisto: Even adding a new function will break the interface, as someone who implements their own `Claims` structure, would need to implement `ValidWithOptions` as well. I think we are stuck with this new feature until v5 :(

gustavosbarreto: > Even adding a new function will break the interface, as someone who implements their own `Claims` structure, would need to implement `ValidWithOptions` as well. I think we are stuck with this new feature until v5 :(

Right. What you think about creating a new interface to extend `Claims`?

```
type Claims interface {
	Valid() bool
}

type ClaimsExtended interface {
	Claims
	ValidWithOptions() bool
}

type RegisteredClaims struct {
}

func (r *RegisteredClaims) Valid() bool {
	fmt.Println("RegisteredClaims::Valid")
	return true
}

type RegisteredClaimsExtended struct {
	RegisteredClaims
}

func (r *RegisteredClaimsExtended) ValidWithOptions() bool {
	fmt.Println("RegisteredClaimsExtended::Valid")
	return true
}

// call function accordingly to interface type (https://github.com/golang-jwt/jwt/blob/1096e506e671d6d6fe134cc997bbd475937392c8/parser.go#L87)
func Verify(claims Claims) bool {
        // is this claims extended?
	if claimsExt, ok := claims.(ClaimsExtended); ok {
		return claimsExt.ValidWithOptions()
	}

	return claims.Valid()
}

func main() {
	claims := &RegisteredClaims{}
	fmt.Println(Verify(claims))

	claimsExt := &RegisteredClaimsExtended{}
	fmt.Println(Verify(claimsExt))
}
```

oxisto (approved):

PR #187, opened 2022-03-27
Chore: Added Github Action to auto format code on push to main

### LINKED ISSUE #93: Add go fmt check on github actions
We have to see what are the available options on formatting, we could either perform an automatic format step upon merge (https://github.com/golang-jwt/jwt/pull/53#issuecomment-892274167) or either add/configure a linter rule for formatting

jay-dee7:
This PR adds a Github Action to auto-format code using `gofumpt` on push to master.
 
Fixes #93 

Signed-off-by: jay-dee7 <jasdeepsingh.uppal@gmail.com>

### DIFF
--- a/.github/workflows/auto-formatter.yml
+++ b/.github/workflows/auto-formatter.yml
@@ -1,0 +1,21 @@
+name: gofumpt
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        ref: ${{ github.head_ref }}
+        fetch-depth: 0
+
+    - name: Auto Format code
+      uses: iamnotaturtle/auto-gofmt@v2.0.0
+      with:
+        only_changed: true


.github/workflows/auto-formatter.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        ref: ${{ github.head_ref }}
+        fetch-depth: 0
+
+    - name: Auto Format code
+      uses: iamnotaturtle/auto-gofmt@v2.0.0
```
mfridman, line 19: Thanks for the PR, but I'd prefer not introducing a 3rd party github action for formatting.

jay-dee7: Hey @mfridman does that mean the regular `goimports` or `go fmt` would be preferred here?

mfridman: Ye, most projects I've seen have linting, formatting, etc. run against the pull request itself. If there is a diff or something doesn't match it is up to the author to fix (CI fails).

I don't have a strong opinion on whether to use `gofmt` or `gofumpt`. The latter has [pre-built binaries](https://github.com/mvdan/gofumpt/releases/tag/v0.3.1) so can curl it.

Something like this can be added before the tests:

```
- name: Check Go code formatting
  run: |
    if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
      gofmt -s -l .
      echo "Please format Go code by running: go fmt ./..."
    exit 1
    fi
```

This avoids a dependency on a GitHub Action and enforces standard PR's to make reviews easier.

jay-dee7: I think the CI failing actually is the way go. I have it failing on all of my other projects. Usually author is responsible to fix them. I'll go ahead and make the PR changes :)

oxisto:
Closed in favour of #206

PR #198, opened 2022-04-27
chore: replace ioutil with io and os

estensen:
`ioutil` was deprecated in [Go 1.16](https://go.dev/doc/go1.16)

### DIFF
--- a/rsa_pss_test.go
+++ b/rsa_pss_test.go
@@ -1,10 +1,11 @@
+//go:build go1.4
 // +build go1.4
 
 package jwt_test
 
 import (
 	"crypto/rsa"
-	"io/ioutil"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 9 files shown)


rsa_pss_test.go
```diff
@@ -1,10 +1,11 @@
+//go:build go1.4
```
mfridman, line 1: Do we still need these build tags? Can we remove them entirely..

mfridman (commented):
Seems reasonable to me. Just a question about the go1.4 build tags.

rsa_pss_test.go
estensen: I think that's out of scope for this PR, but I'm happy to create another PR removing them. This build line is added because [Go 1.17](https://go.dev/doc/go1.17) has a new syntax for them.

mfridman: That's fair.

mfridman (approved):

mfridman:
cc @golang-jwt/maintainers what are your thoughts?

oxisto:
> cc @golang-jwt/maintainers what are your thoughts?

Fine by me.

oxisto (approved):

PR #200, opened 2022-05-02
Fixed integer overflow in NumericDate.MarshalJSON

### LINKED ISSUE #199: NumericDate.MarshalJSON overflows on large unix timestamps
Marshalling `NewNumericDate(time.Unix(253402271999, 0))` will result in `-4852145033` instead of `253402271999`

Updates:
This is due to the fact that `Time.UnixNano()` overflows, and `NumericDate.MarshalJSON()` internally uses this method.

PR #200 created to workaround this std library limitation.

qqiao:
Fixed #199 

The original issue was caused by the fact that if we use a very large
unix timestamp. The resulting value from `Time.UnixNano()`
overflows a int64, as documented here: https://pkg.go.dev/time#Time.UnixNano.

This patch works around the issue by treating the second part and
nanosecond part separately, taking the second part from `Time.Unix()`,
and the nanosecond part from `Time.Nanosecond()`, formatting them
separately, and then combining the resulting strings.

This approach allows us to correctly marshal all go `time.Time` instances,
whereas using `Time.Unix` won't give us enough precision, and either
`Time.UnixMicro` or `Time.UnixNano` would overflow at some point for
even valid `time.Time` instances.

The added benefit of this approach is that we no longer have to deal
with the added nanosecond delta that Go uses in order to achieve
monotonic time, so the result of marshaling `time.Unix(0, 123456)` will
simply be `0.123456` instead of `0.123456 + small_delta`.

### DIFF
--- a/types.go
+++ b/types.go
@@ -53,12 +53,22 @@ func (date NumericDate) MarshalJSON() (b []byte, err error) {
 	if TimePrecision < time.Second {
 		prec = int(math.Log10(float64(time.Second) / float64(TimePrecision)))
 	}
-	trancatedDate := date.Truncate(TimePrecision)
-	whole := strconv.FormatInt(trancatedDate.Unix(), 10)
-	decimal := strconv.FormatFloat(float64(trancatedDate.Nanosecond())/float64(time.Second), 'f', prec, 64)
-	f := append([]byte(whole), []byte(decimal)[1:]...)
+	truncatedDate := date.Truncate(TimePrecision)
 
-	return f, nil
+	// For very large timestamps, UnixNano would overflow an int64, but this
+	// function requires nanosecond level precision, so we have to use the
+	// following technique to get round the issue:
+	// 1. Take the normal unix timestamp to form the seconds number part,
+	// 2. Take the result of the Nanosecond function, which retuns the offset
+	//    within the second of the particular unix time instance, to form the
+	//    decimal part of the result
+	// 3. Concatenate the seconds and the decimal part to produce the final result
+	seconds := strconv.FormatInt(truncatedDate.Unix(), 10)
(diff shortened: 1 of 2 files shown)


types.go
```diff
@@ -53,9 +53,12 @@ func (date NumericDate) MarshalJSON() (b []byte, err error) {
 	if TimePrecision < time.Second {
 		prec = int(math.Log10(float64(time.Second) / float64(TimePrecision)))
 	}
-	f := float64(date.Truncate(TimePrecision).UnixNano()) / float64(time.Second)
+	trancatedDate := date.Truncate(TimePrecision)
```
oxisto, line 56: `truncatedDate`

types.go
```diff
@@ -53,9 +53,12 @@ func (date NumericDate) MarshalJSON() (b []byte, err error) {
...
+	whole := strconv.FormatInt(trancatedDate.Unix(), 10)
```
oxisto, line 57: can we come up with a better variable name than `whole`?

types.go
```diff
@@ -53,9 +53,12 @@ func (date NumericDate) MarshalJSON() (b []byte, err error) {
...
+	decimal := strconv.FormatFloat(float64(trancatedDate.Nanosecond())/float64(time.Second), 'f', prec, 64)
+	f := append([]byte(whole), []byte(decimal)[1:]...)
```
oxisto, line 59: I think a small comment to the whole block would be good, in order to understand what is happening here.

oxisto (commented):
Sorry for the delay on this. I was quite busy for personal reasons for a while. Looks good overall, just some minor details like variable naming and some comments would be nice.

types.go
qqiao: Fixed

types.go
qqiao: Done

oxisto (approved):
Thanks for the documentation.

mfridman (approved):
Sorry for the delay, thanks for submitting this. LGTM.

PR #207, opened 2022-05-28
Update SECURITY.md

oxisto:
Added mailing list for disclosing security issues.


mfridman:
How does it work in terms of who is cc'd on this mailing list (and whether it's private)? Sorry for the naive question 😓

oxisto:
You should have received an email from the Google groups. Currently only we two are on the list and only members can view. Non-members can write but it’s moderated.

mfridman (approved):
thanks for putting that together.

PR #208, opened 2022-05-28
Validation Options - Experiment 1: Embedding validation options in claim

oxisto:

Option 1 is to embedded a `validator` struct which olds all the necessary information into the claims itself. It is populated by the Parser during `Parse`. The claim's `Valid` function can then retrieve the information and adjust the validation.

Pros:
* Keeps everything as is on the outside, no changes to the exported `Valid` or `VerifyXXX` functions, not even with varargs.
* Pretty simple to use for the user I guess
* The variable is unexported in the claims itself (`v`) and should therefore not interfere with json Marshalling

Cons:
* can only work with `jwt.RegisteredClaims` (and jwt.StandardClaims) because I have no way to "silently" inject it into `MapClaims`. I could add a key to the map, but this will be then transparent to the user and might conflict with a custom claim the user has
*  The way the current test is structured in `parser_test.go` I cannot really test it because it is using the `jwt_test` package for tests and therefore cannot access the unexpected `v` field and this leads to some errors in the comparison test.

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -22,6 +22,8 @@ type Claims interface {
 //
 // See examples for how to use this with your own claim types.
 type RegisteredClaims struct {
+	v validator
+
 	// the `iss` (Issuer) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1
 	Issuer string `json:"iss,omitempty"`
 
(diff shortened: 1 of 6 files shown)


oxisto:

PR #210, opened 2022-05-28
Validation Options - Experiment 3 : Introducing new (optional) `ClaimsWithOptions` interface next to `Claims`

oxisto:

Option 3 is to create a new `ClaimsWithOptions` interface next to `Claims`, which provides an additional `ValidWithOptions` function in the claims to check validations with options. `Valid` is then redirected to `ValidWithOptions`

Pros:
* Should not break anything and seems to be transparent to the user

Cons:
* It's kind of a mess behind the scenes and it introduces a new interface that is sort of a workaround that we might want to get rid of in v5 again if we decide to break the api and just supply options to `Valid` directly.

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -9,7 +9,27 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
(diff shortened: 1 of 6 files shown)


oxisto:

PR #215, opened 2022-06-08
chore: remove unused claims in table driven test - part2

gkech:
Followup PR: https://github.com/golang-jwt/jwt/pull/212

Noticed that `claims` are not used in various table-driven tests around the project. With this PR we clean them up. Sorry for not noticing in https://github.com/golang-jwt/jwt/pull/212.

### DIFF
--- a/ecdsa_test.go
+++ b/ecdsa_test.go
@@ -14,7 +14,6 @@ var ecdsaTestData = []struct {
 	keys        map[string]string
 	tokenString string
 	alg         string
-	claims      map[string]interface{}
 	valid       bool
 }{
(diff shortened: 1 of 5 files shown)


oxisto:
Wouldn't it make more sense to actually *check* the claims instead of removing them?

gkech:
> Wouldn't it make more sense to actually _check_ the claims instead of removing them?

Hello @oxisto, thank you for your comment. 

Initially, I was thinking the same but I wasn't sure if the lack of claims checking was intentional e.g. because they are covered in another unit test.

Issue #67, opened 2021-08-03
Add support for encryption (RFC7516)


lggomez:
Migrated from https://github.com/dgrijalva/jwt-go/issues/346:

> It would be great if this library could also support payload encryption (https://tools.ietf.org/html/rfc7516)
> 
> It would greatly help in case sensitive information needs to be stored inside the JWT token, and offload the need to encrypt it in every application separately

lggomez:
>  dgrijalva commented on Jul 30, 2019
> 
> I've looked into it. JWE is a sibling standard to JWT. I'm open to supporting it, but it's considerably more complex, so it'll require some work and an extended API. At the moment, I'm trying to wrap up the next major version. It would be best to take a look at JWE again after that ships.

ptman:
JWE support seems to still be missing. What other libraries have people had good success with?

PR #220, opened 2022-06-23
JWE Support

ortyomka:
Issue #67 

Add support of JWE with one key (Compact)

Add AES GCM cipher to encrypt content
Add RSA-OAEP to encrypt key

Add test from RFC7516 [Section 3.3](https://datatracker.ietf.org/doc/html/rfc7516#section-3.3) 

cc @mfridman

### DIFF
--- a/jwe/aesgcm.go
+++ b/jwe/aesgcm.go
@@ -3,8 +3,8 @@ package jwe
 import (
 	"crypto/aes"
 	"crypto/cipher"
-	"crypto/rand"
 	"errors"
+	"io"
 )
 
 var (
@@ -18,7 +18,7 @@ const TagSizeAESGCM = 16
 
 type EncryptionType string
 
-var A256GCM = EncryptionType("A256GCM")
+var EncryptionTypeA256GCM = EncryptionType("A256GCM")
 
 type cipherAESGCM struct {
 	keySize int
@@ -36,7 +36,7 @@ func (ci cipherAESGCM) encrypt(key, aad, plaintext []byte) (iv []byte, ciphertex
 	}
 
 	iv = make([]byte, aead.NonceSize())
-	_, err = rand.Read(iv)
+	_, err = io.ReadFull(RandReader, iv)
 	if err != nil {
 		return nil, nil, nil, err
 	}
@@ -84,7 +84,7 @@ func newAESGCM(keySize int) *cipherAESGCM {
 
 func getCipher(alg EncryptionType) (*cipherAESGCM, error) {
 	switch alg {
-	case A256GCM:
+	case EncryptionTypeA256GCM:
 		return newAESGCM(32), nil
 	default:
 		return nil, ErrUnsupportedEncryptionType
--- a/jwe/aesgcm.go
+++ b/jwe/aesgcm.go
@@ -1,0 +1,92 @@
+package jwe
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+)
+
+var (
+	ErrInvalidKeySize            = errors.New("invalid key size")
+	ErrInvalidTagSize            = errors.New("invalid tag size")
+	ErrInvalidNonceSize          = errors.New("invalid nonce size")
+	ErrUnsupportedEncryptionType = errors.New("unsupported encryption type")
+)
+
+const TagSizeAESGCM = 16
+
+type EncryptionType string
+
+var A256GCM = EncryptionType("A256GCM")
+
+type cipherAESGCM struct {
+	keySize int
+	getAEAD func(key []byte) (cipher.AEAD, error)
+}
+
+func (ci cipherAESGCM) encrypt(key, aad, plaintext []byte) (iv []byte, ciphertext []byte, tag []byte, err error) {
+	if len(key) != ci.keySize {
+		return nil, nil, nil, ErrInvalidKeySize
+	}
+
+	aead, err := ci.getAEAD(key)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	iv = make([]byte, aead.NonceSize())
+	_, err = rand.Read(iv)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	res := aead.Seal(nil, iv, plaintext, aad)
+	tagIndex := len(res) - TagSizeAESGCM
+
+	return iv, res[:tagIndex], res[tagIndex:], nil
+}
+
+func (ci cipherAESGCM) decrypt(key, aad, iv []byte, ciphertext []byte, tag []byte) ([]byte, error) {
+	if len(key) != ci.keySize {
+		return nil, ErrInvalidKeySize
+	}
+
+	if len(tag) != TagSizeAESGCM {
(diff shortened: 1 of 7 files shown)


mfridman:
I think this was an oversight on our part. This likely doesn't belong in this package.

https://github.com/golang-jwt/jwt#signing-vs-encryption

Not sure if our stance on this has changed cc @oxisto

oxisto:
> I think this was an oversight on our part. This likely doesn't belong in this package.
> 
> https://github.com/golang-jwt/jwt#signing-vs-encryption
> 
> Not sure if our stance on this has changed cc @oxisto

Personally, I wouldn’t mind JWE support. We probably need to come up with a good naming / packaging scheme. I hopefully can allocate some time next week to give this a first review round.

ortyomka:
Reminder about PR

@oxisto @mfridman

oxisto:
Before we go any further with this, I wonder whether it would be more appropriate to move this into a separate repository, @mfridman? Since JWT and JWE are more sibling references, and JWE support within the JWT repository might be confusing. So we would have `github.com/golang-jwt/jwt` and `github.com/golang-jwt/jwe`.

jwe/aesgcm.go
```diff
@@ -0,0 +1,92 @@
...
+	ErrInvalidTagSize            = errors.New("invalid tag size")
+	ErrInvalidNonceSize          = errors.New("invalid nonce size")
+	ErrUnsupportedEncryptionType = errors.New("unsupported encryption type")
+)
+
+const TagSizeAESGCM = 16
+
+type EncryptionType string
+
+var A256GCM = EncryptionType("A256GCM")
```
oxisto, line 21: Should probably be EncryptionTypeA256GCM

jwe/jwe.go
```diff
@@ -0,0 +1,69 @@
+package jwe
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+func NewJWE(alg KeyAlgorithm, key interface{}, method EncryptionType, plaintext []byte) (*jwe, error) {
```
oxisto, line 9: Documentation for exported function

jwe/jwe.go
```diff
@@ -0,0 +1,69 @@
...
+	jwe.iv, jwe.ciphertext, jwe.tag, err = chipher.encrypt(cek, []byte(rawProtectedBase64), plaintext)
+	if err != nil {
+		return nil, err
+	}
+
+	return jwe, nil
+}
+
+type jwe struct {
+	protected    map[string]string
```
oxisto, line 49: Would it make sense to make this header a struct with specific values?

jwe/jwe_test.go
```diff
@@ -0,0 +1,56 @@
+package jwe_test
```
oxisto, line 1: It would be nice if we could re-use example data from https://datatracker.ietf.org/doc/html/rfc7516#appendix-A

jwe/jwe.go
oxisto, line 9: Plaintext is the JWT in compact form, correct?

oxisto (commented):
First round of reviews

jwe/keycrypter.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	ErrUnsupportedKeyType      = errors.New("unsupported key type")
+	ErrUnsupportedKeyAlgorithm = errors.New("unsupported key algorithm")
+)
+
+type KeyAlgorithm string
+
+var RSA_OAEP = KeyAlgorithm("RSA-OAEP")
```
oxisto, line 17: KeyAlgorithmRSAOAEP

jwe/keycrypter.go
```diff
@@ -0,0 +1,53 @@
...
+
+func rsaEncrypt(key *rsa.PublicKey, cek []byte, alg KeyAlgorithm) ([]byte, error) {
```
oxisto, line 19: any particular reason why you are using functions here instead of a struct with methods? Just wondering. The latter might be better if we want to make an interface out of it.

jwe/aesgcm.go
ortyomka: Renamed

jwe/jwe.go
ortyomka: Added documentation

jwe/jwe.go
ortyomka: Changed to struct

jwe/jwe.go
ortyomka: Any sensitive info. It can be JWT in compact form.

jwe/keycrypter.go
ortyomka: Renamed

jwe/keycrypter.go
ortyomka: Changed to struct with Encrypt and Decrypt. For the future

ortyomka:
@oxisto thank you for the review. I fixed all comments, ask for re-review.

mfridman:
> Before we go any further with this, I wonder whether it would be more appropriate to move this into a separate repository, @mfridman? Since JWT and JWE are more sibling references, and JWE support within the JWT repository might be confusing. So we would have `github.com/golang-jwt/jwt` and `github.com/golang-jwt/jwe`.

Yes, I think this should be moved to its own repository. `github.com/golang-jwt/jwe` sounds fine to me.

ortyomka:
> > Before we go any further with this, I wonder whether it would be more appropriate to move this into a separate repository, @mfridman? Since JWT and JWE are more sibling references, and JWE support within the JWT repository might be confusing. So we would have `github.com/golang-jwt/jwt` and `github.com/golang-jwt/jwe`.
> 
> Yes, I think this should be moved to its own repository. `github.com/golang-jwt/jwe` sounds fine to me.

@mfridman, who can create new repo? I will forward this PR to there

oxisto:
> > > Before we go any further with this, I wonder whether it would be more appropriate to move this into a separate repository, @mfridman? Since JWT and JWE are more sibling references, and JWE support within the JWT repository might be confusing. So we would have `github.com/golang-jwt/jwt` and `github.com/golang-jwt/jwe`.
> > 
> > 
> > Yes, I think this should be moved to its own repository. `github.com/golang-jwt/jwe` sounds fine to me.
> 
> @mfridman, who can create new repo? I will forward this PR to there

https://github.com/golang-jwt/jwe

ortyomka:
> > > > Before we go any further with this, I wonder whether it would be more appropriate to move this into a separate repository, @mfridman? Since JWT and JWE are more sibling references, and JWE support within the JWT repository might be confusing. So we would have `github.com/golang-jwt/jwt` and `github.com/golang-jwt/jwe`.
> > > 
> > > 
> > > Yes, I think this should be moved to its own repository. `github.com/golang-jwt/jwe` sounds fine to me.
> > 
> > 
> > @mfridman, who can create new repo? I will forward this PR to there
> 
> https://github.com/golang-jwt/jwe

@oxisto need some init commit, I cannot fork or open PR

ortyomka:
Moved to https://github.com/golang-jwt/jwe/pull/1

PR #221, opened 2022-06-26
Fix typo in curve bits 512 vs 521

elarivie:


oxisto:
That is not a typo. ES512 uses P-521 and SHA-512. See https://datatracker.ietf.org/doc/html/rfc7518#section-3.1

PR #226, opened 2022-08-10
Implement a BearerExtractor

WhyNotHugo:
This is a rather common extractor; it extracts the JWT from the HTTP
Authorization header, expecting it to include the "Bearer " prefix.

This patterns is rather common and this snippet is repeated in enough
applications that it's probably best to just include it upstream and
allow reusing it.

### DIFF
--- a/request/extractor.go
+++ b/request/extractor.go
@@ -3,6 +3,7 @@ package request
 import (
 	"errors"
 	"net/http"
+	"strings"
 )
 
 // Errors
@@ -78,4 +79,17 @@ func (e *PostExtractionFilter) ExtractToken(req *http.Request) (string, error) {
 	} else {
 		return "", err
 	}
+}
+
+// BearerExtractor extracts a token from the Authorization header.
+// The header is expected to match the format "Bearer XX", where "XX" is the
+// JWT token.
+type BearerExtractor struct{}
+
+func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
+	tokenHeader := req.Header.Get("Authorization")
+	if tokenHeader == "" || !strings.HasPrefix(tokenHeader, "Bearer ") {
+		return "", ErrNoTokenInRequest
+	}
+	return strings.TrimPrefix(tokenHeader, "Bearer "), nil
 }
--- a/request/extractor.go
+++ b/request/extractor.go
@@ -88,8 +88,10 @@ type BearerExtractor struct{}
 
 func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
 	tokenHeader := req.Header.Get("Authorization")
-	if tokenHeader == "" || !strings.HasPrefix(tokenHeader, "Bearer ") {
+	// The usual convention is for "Bearer" to be title-cased. However, there's no
+	// strict rule around this, and it's best to follow the robustness principle here.
+	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader), "bearer ") {
 		return "", ErrNoTokenInRequest
 	}
-	return strings.TrimPrefix(tokenHeader, "Bearer "), nil
+	return tokenHeader[7:], nil
 }
--- a/request/extractor_test.go
+++ b/request/extractor_test.go
@@ -91,16 +91,21 @@ func makeExampleRequest(method, path string, headers map[string]string, urlArgs 
 }
 
 func TestBearerExtractor(t *testing.T) {
-	request := makeExampleRequest("POST", "https://example.com/", map[string]string{"Authorization": "Bearer 123"}, nil)
+	request := makeExampleRequest("POST", "https://example.com/", map[string]string{"Authorization": "Bearer ToKen"}, nil)
 	token, err := BearerExtractor{}.ExtractToken(request)
-	if err != nil || token != "123" {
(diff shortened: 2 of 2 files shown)


oxisto:
Good idea. May I suggest an alternative name like `AuthorizationHeaderExtractor` or do you think that is too long?

mfridman:
I have this in literally every project, haha. I'd suggest something like `GetTokenFromHeader`

oxisto:
> I have this in literally every project, haha. I'd suggest something like `GetTokenFromHeader`

There is already a `TokenExtractor` which loops through all tokens, since this is specific to a bearer token we might as well stick with the original name.

request/extractor.go
```diff
@@ -79,3 +80,16 @@ func (e *PostExtractionFilter) ExtractToken(req *http.Request) (string, error) {
...
 }
+
+// BearerExtractor extracts a token from the Authorization header.
+// The header is expected to match the format "Bearer XX", where "XX" is the
+// JWT token.
+type BearerExtractor struct{}
+
+func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
+	tokenHeader := req.Header.Get("Authorization")
+	if tokenHeader == "" || !strings.HasPrefix(tokenHeader, "Bearer ") {
```
oxisto, line 91: You might consider making this case insensitive. The headers are definitely case insensitive, but Headers.Get already takes care of this. I need to Look it up again but i am pretty sure `bearer` is also ok.

oxisto: I tried to find the actual RFC that specifies this, which is... let's say complicated. There are some who think that it is insensitive (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#directives). However, it is not really 100 % specified in the RFC. It is only specified in the `WWW-Authenticate` server response (to be case-insensitive), so I would keep this as is for now.

Unless some RFC expert can shed some more insight on this.

oxisto (approved):

request/extractor.go
WhyNotHugo: Updates this to be case-insensitive. It's best to be flexible here.

request/extractor.go
```diff
@@ -79,3 +80,18 @@ func (e *PostExtractionFilter) ExtractToken(req *http.Request) (string, error) {
...
+type BearerExtractor struct{}
+
+func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
+	tokenHeader := req.Header.Get("Authorization")
+	// The usual convention is for "Bearer" to be title-cased. However, there's no
+	// strict rule around this, and it's best to follow the robustness principle here.
+	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader), "bearer ") {
+		return "", ErrNoTokenInRequest
+	}
+	return strings.TrimPrefix(tokenHeader, "Bearer "), nil
```
oxisto, line 96:
```suggestion
	return strings.TrimPrefix(strings.ToLower(tokenHeader), "bearer "), nil
```

WhyNotHugo: This will convert the token into lowercase. Are tokens themselves case-insensitive too?

request/extractor.go
```diff
@@ -79,3 +80,18 @@ func (e *PostExtractionFilter) ExtractToken(req *http.Request) (string, error) {
 		return "", err
 	}
 }
+
+// BearerExtractor extracts a token from the Authorization header.
+// The header is expected to match the format "Bearer XX", where "XX" is the
+// JWT token.
+type BearerExtractor struct{}
```
mfridman, line 87: I might have missed it, but why does this need an extra struct? Wouldn't a single function that takes `*http.Request` and returns `string, error` suffice?

mfridman: I assume there is an interface specific to your code base, but I'm not sure this make sense without also defining the interface within this codebase (which I'd rather not do).

request/extractor.go
oxisto: Oops. No they are not

request/extractor.go
oxisto: It follows the extractor interface here https://github.com/golang-jwt/jwt/blob/61ab9b21f8e1ddb45e0f7bf340d0d874408a4cc7/request/extractor.go#L17

WhyNotHugo: See [github.com/dgrijalva/jwt-go/v4/request#ParseFromRequest](https://godocs.io/github.com/dgrijalva/jwt-go/v4/request#ParseFromRequest).

request/extractor.go
WhyNotHugo: Updated PR to avoid messing up case of the token.

WhyNotHugo: And added tests for that.

request/extractor.go
WhyNotHugo: Also tests cover this now.

oxisto (approved):

mfridman (approved):

AlexanderYastrebov:
There is a long-existing `AuthorizationHeaderExtractor` that does the same (almost, as it also allows token without `Bearer ` prefix afaict)
https://github.com/golang-jwt/jwt/blob/2ebb50f957d606de5909fcf9ed49f9af3bc35e97/request/oauth2.go#L7-L21

PR #233, opened 2022-08-26
fix: SignedString/ParseWithClaims method must byte array

thoohv5:
SignedString/ParseWithClaims params need interface{}, but In fact need byte array, so conversion interface to byte array

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,0 +1,59 @@
+package jwt
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// ByteArrayVal interface to byte array
+func ByteArrayVal(value interface{}) (key []byte) {
+	if value == nil {
+		return
+	}
+	switch value.(type) {
+	case float64:
+		ft := value.(float64)
+		key = []byte(strconv.FormatFloat(ft, 'f', -1, 64))
+	case float32:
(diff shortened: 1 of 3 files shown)


util.go
```diff
@@ -0,0 +1,59 @@
...
+	"encoding/json"
+	"strconv"
+)
+
+// ByteArrayVal interface to byte array
+func ByteArrayVal(value interface{}) (key []byte) {
+	if value == nil {
+		return
+	}
+	switch value.(type) {
```
eduardo-mior, line 13: Instead of doing the CAST within each SWITCH CASE you can create a variable already typed.


Example:
```golang
switch typedValue := value.(type) {
....
}
```

Official documentation: https://go.dev/tour/methods/16

oxisto:
The type needed depends on the signature method you are using. If you are using a symmetric one, it needs to be `byte[]', but for asymmetric method, such as RSA it needs to be a public/private key. You can find a list here: https://github.com/golang-jwt/jwt#signing-methods-and-key-types

The parameter needs to be an interface{} and each signature method is then responsible for checking if it is the correct type. Possibly in the future, we can solve this better with Go generics.

PR #234, opened 2022-08-27
`v5` Pre-Release

oxisto:
This commit serves as the basis for further `v5` developments. It will introduce some API-breaking changes, especially to the way tokens are validated. This will allow us to provide some long-wanted features with regards to the validation API. We are aiming to do this as smoothly as possible, however, with any major version. please expect that you might need to adapt your code.

The actual development will be done in the course of the next week, if time permits. It will be done in seperate PRs that will use this PR as a base. Afterwards, we will probably merge this and release an initial 5.0.0-alpha1 or similar.

`v5` Checklist:

- [x] New validation API
- [x] Cleanup / harmonize error messages
- [x] Increase code coverage


mfridman:
@oxisto Apologies, I took some time off end of December and just ran out of steam. I'd still like to work through this and get to a stable `/v5`, is this the right place to start reviewing?

We could also create a milestone with all the PR's and/or issues attached to it, helps differentiate with the rest of the noise.

oxisto:
> @oxisto Apologies, I took some time off end of December and just ran out of steam. I'd still like to work through this and get to a stable `/v5`, is this the right place to start reviewing?

Yes

> We could also create a milestone with all the PR's and/or issues attached to it, helps differentiate with the rest of the noise.

there is a label “next” that I used to specify which PRs should end up in `v5`. I think we wanted to add more test coverage but from the features we are pretty much complete I would say.

go.mod
```diff
@@ -1,7 +1,3 @@
-module github.com/golang-jwt/jwt/v4
+module github.com/golang-jwt/jwt/v5
```
mfridman, line 1: In order for users to continue fetching the /v4 module, and for us to patch issues in the v4.x.y version .. should we also have a `v4` branch that branches off from 9358574a7a1a2c8d644f22b6e8de627ba85c58d0 (or whatever the last commit before they diverge)?

oxisto: We can probably do that, yes

mfridman: I published a `v4.5.0` tag for 9358574a and created a separate [v4 branch ](https://github.com/golang-jwt/jwt/tree/v4).

claims.go
```diff
@@ -1,269 +1,16 @@
...
-)
-
-// Claims must just have a Valid method that determines
-// if the token is invalid for any supported reason
+// Claims represent any form of a JWT Claims Set according to
+// https://datatracker.ietf.org/doc/html/rfc7519#section-4. In order to have a
+// common basis for validation, it is required that an implementation is able to
+// supply at least the claim names provided in
+// https://datatracker.ietf.org/doc/html/rfc7519#section-4.1 namely `exp`,
+// `iat`, `nbf`, `iss` and `aud`.
```
mfridman, line 8:
```suggestion
// `iat`, `nbf`, `iss`, `sub` and `aud`.
```

map_claims.go
```diff
@@ -2,150 +2,107 @@ package jwt
...
-		for _, a := range v {
-			vs, ok := a.(string)
-			if !ok {
-				return false
-			}
-			aud = append(aud, vs)
-		}
-	}
-	return verifyAud(aud, cmp, req)
+// GetExpirationTime implements the Claims interface.
```
mfridman, line 11: Above this line could add something like:

```go
var _ Claims = (MapClaims)(nil)
```

Makes it a bit easier to reason that this type is intended to satisfy the `Claims` interface. And then can get rid of the "... implements the Claims interface." comments on the methods.

map_claims.go
```diff
@@ -2,150 +2,107 @@ package jwt
...
+// GetSubject implements the Claims interface.
+func (m MapClaims) GetSubject() (string, error) {
+	return m.ParseString("sub")
+}
 
-	v, ok := m["iat"]
+// ParseNumericDate tries to parse a key in the map claims type as a number
+// date. This will succeed, if the underlying type is either a [float64] or a
+// [json.Number]. Otherwise, nil will be returned.
+func (m MapClaims) ParseNumericDate(key string) (*NumericDate, error) {
```
mfridman, line 44:
```suggestion
func (m MapClaims) parseNumericDate(key string) (*NumericDate, error) {
```

Can this (and the other helpers) be unexported, feels like an internal implementation detail .. we can always export them if needed.

mfridman: Reading the helpers a bit more, I think we definitely don't want to expose these because there is unexpected behaviour. E.g.,

```
v, ok := m[key]
if !ok {
	return nil, nil
}
```

Should this return an `ErrClaimNotFound` error?

validator.go
```diff
@@ -0,0 +1,261 @@
...
+	expectedIss string
+
+	// expectedSub contains the subject this token expects. Supplying an empty
+	// string will disable sub checking.
+	expectedSub string
+}
+
+// CustomClaims represents a custom claims interface, which can be built upon the integrated
+// claim types, such as map claims or registered claims.
+type CustomClaims interface {
```
mfridman, line 43: I'm struggling a bit with this one. Can we ship `v5` without this interface, and consider this in minor release?

The caller already has access to the claims, and they can do whatever validation they need. Why do we need an extra interface only to call the `CustomValidation()` method on behalf of the caller.

validator.go
```diff
@@ -0,0 +1,261 @@
...
+
+// VerifyExpiresAt compares the exp claim in claims against cmp. This function
+// will return true if cmp < exp. Additional leeway is taken into account.
+//
+// If exp is not set, it will return true if the claim is not required,
+// otherwise false will be returned.
+//
+// Additionally, if any error occurs while retrieving the claim, e.g., when its
+// the wrong type, false will be returned.
+func (v *validator) VerifyExpiresAt(claims Claims, cmp time.Time, required bool) bool {
```
mfridman, line 132: Not sure if it's a good idea, but since we're here already .. wdyt about these helpers returning an `error` and removing the `required bool` arg?

validator.go
```diff
@@ -0,0 +1,261 @@
...
+		now = v.timeFunc()
+	} else {
+		now = time.Now()
+	}
+
+	// We always need to check the expiration time, but usage of the claim
+	// itself is OPTIONAL
+	if !v.VerifyExpiresAt(claims, now, false) {
+		vErr.Inner = ErrTokenExpired
+		vErr.Errors |= ValidationErrorExpired
```
mfridman, line 73: I know I asked about this before, but can we remove these operators in favor of plain ol' errors?

map_claims.go
oxisto: The primary reason was just to have a piece of documentation in the individual functions to understand where they are coming from.

validator.go
oxisto: This was sort of a compromise since my feeling was that currently a lot of people are overriding the `Valid` method on their custom claim to add additional piece of validation in the "correct" place. Since claims do not have the valid method anymore, I wanted users to give some "canonical" place to do additional validation. I believe this is a valid use case and much cleaner than first doing the "regular" claims validation and then if that succeeds doing an additional one, with extra error handling, etc.

Having this here in the example is quite nice imho.
https://github.com/golang-jwt/jwt/blob/3a9ee81ba34953206aea4c4ebbd5d78fa2afa339/example_test.go#L119-L130

The question is, whether this interface could actually be unexported, so we are more flexible with naming it, etc.?

validator.go
oxisto: Yes and no :) We could for example replace `Errors` with a slice or `error` instead of the bitfield. I could then adjust the `errors.Is` to work with that. Once we only support Go 1.20+ we could even make use of the new `errors.Join` function, and `Unwrap` of multiple errors (see https://tip.golang.org/doc/go1.20#errors).

I can try to play around with that in a separate PR that merges into the branch.

map_claims.go
oxisto: That depends on where you want to catch the "non-existence". They way it is built is that this function returns a `nil` numeric date if the date field is not set. This is not really an error, because the date fields such as `exp` and such are optional. However, if the field is populated with a non-date type, then an error must be thrown, otherwise this could lead to problems in the token verification step.

map_claims.go
```diff
@@ -2,150 +2,107 @@ package jwt
...
+		return nil, nil
 	}
 
-	switch iat := v.(type) {
+	switch exp := v.(type) {
 	case float64:
-		if iat == 0 {
-			return verifyIat(nil, cmpTime, req)
+		if exp == 0 {
+			return nil, nil
```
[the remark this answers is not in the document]
oxisto: I basically replicated the behaviour of the old code here. I think, similar to the old code `0` here means basically an "unset" float value and that corresponds to a `nil` date. We could however just also create a "zero" date from it and pass it to `newNumericDateFromSeconds` like every other value (not quite sure why we do not do that). It is definitely not an error though.

map_claims.go
oxisto: I can unexpert them for now

validator.go
oxisto: I would keep the required, because we might want to give some control over which claims are required or not; this might be application specific. But we could definitely directly return the corresponding error here instead of catching the bool and assigning the error in `Validate` (see your comment below).

validator.go
oxisto: I played around with the idea and this is what I came up with: https://github.com/golang-jwt/jwt/pull/274.

However that uses/needs Go 1.20 for the new fancy multiple error wrapping (which is actually *exactly* what we need). While we had this discussion before, I personally could live with the idea of the `v5` branch "starting" at Go 1.20. Otherwise we could probably have some legacy layer that replicates the behaviour and then switch to the native functions at some point.

validator.go
mfridman: My worry is this is a potential footgun for users because the indirection of second interface.

If users are embedding custom claims, one could argue it should be their responsibuility to validate those claims. But I do see the argument for having a canonical place to do validation.

My general thinking around this package (I could be wrong) is to do the bare minimum to verify the token and validate the claims with an API that eliminates as many footguns as possible.

mfridman: Another option could be to inline the type?

```go
// Finally, we want to give the claim itself some possibility to do some
// additional custom validation based on a custom function
type customValidator interface {
	Validate() error
}
cvt, ok := claims.(customValidator)
if ok {
	if err := cvt.Validate(); err != nil {
		vErr.Inner = err
		vErr.Errors |= ValidationErrorClaimsInvalid
	}
}
```

oxisto: Maybe I am missing the point about your "indirection". Are you concerned about the naming? The way you did it in the example also looks good to me.

oxisto: See https://github.com/golang-jwt/jwt/pull/234/commits/1e16f55059b4ef3ce614e7d5b22bc347cb253ef3

mfridman: I was going back and forth on how to best expose this to users. After thinking about it some more, it feels like we should expose this interface and document its use-case, the way you had it makes it discoverable and allows a compile-time assertion of a custom claims.

```go
type CustomValidator interface {
	Validate() error
}
```

Wdyt?

validator.go
mfridman: Thanks for taking a stab at this, after a bit more thought I suggest we leave this as-is (for now). The most important thing is users don't actually have to do the bit manipulation, and can use `errors` to assert for the correct errors.

I think once Go reaches v1.21 or higher we can revisit the errors story (hopefully it'll just be an implemntation detail).

validator.go
oxisto: Either one is fine for me. Lets keep it inlined for now (for the release candidate) and document its use case in the upcoming docs.

validator.go
oxisto: > Thanks for taking a stab at this, after a bit more thought I suggest we leave this as-is (for now). The most important thing is users don't actually have to do the bit manipulation, and can use `errors` to assert for the correct errors.

Yes that can. 

> 
> I think once Go reaches v1.21 or higher we can revisit the errors story (hopefully it'll just be an implemntation detail).

Although a new major version might be a good point to make such changes and I actually quite like the very clean approach it could offer in terms of errors.

I will see whether I can play around with "simulating" the Go 1.20 behaviour on Go < 1.20 by basically keeping an (unexported) `validationError` (which acts as a compatibility layer) and once we have +2 release versions, we can just go to Go 1.20 as min version and drop the compatibility layer.

validator.go
mfridman: The ergonomics of these methods always felt a bit off, but that could just be me. I wish we didn't expose these to end-users because it swallows errors and there are 3 ways to do things:

- calling `.Validate()` on `*validator` (all)
- calling `.Verify()` and friends on `*validator` (individually)
- calling a custom `Validate() error` (custom)

I presume these Verify.. helper methods exist so users don't have to write the verification logic themselves? 

Fwiw I don't feel strongly about this, but given this is an opportunity to clean some things up I figured it was worth dicussing.

validator.go
mfridman: Ugh, on the other hand this feels a bit confusing, because we have a `(v *validator) Validate(...)` method that takes a `Claims` interface .. and now that interface needs to satisfy a `CustomValidator` interface with a `Validate() error` method.

Users need to create a concrete type, that embeds a type or implements one that satisfies a `Claims` interface. And then it also needs to have a `Validate() error` to satisfy a `CustomValidator` interface.

Is this complicated, and can it be made easier. That's kind of what I'm trying to explore.

mfridman: I tried this out, and this interface helps glue together the claims in the package and any custom validation, maybe this is the least-worst option?

```go
type MyCustomClaims struct {
	Foo string `json:"foo"`
	jwt.RegisteredClaims
}

var _ jwt.ClaimsValidator = (*MyCustomClaims)(nil)

func (m MyCustomClaims) Validate() error {
	if m.Foo != "bar" {
		return errors.New("must be foobar")
	}
	return nil
}
```

validator.go
mfridman: I presume we could have 2 branches of code guarded by build tags? Although this may be more effort than it's worth.

If it comes down to that, I can get behind the `/v5` version being >=v1.20 minimum version

map_claims.go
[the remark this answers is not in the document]
mfridman: > We could however just also create a "zero" date from it and pass it to newNumericDateFromSeconds like every other value

That's what I would have expected, might save someone a panic down the road. Although in all fairness, maybe a `nil` date here is fine (many folks struggle working with dates), and we should document that the caller is responsible for checking both return values.

validator.go
oxisto: > I presume we could have 2 branches of code guarded by build tags? Although this may be more effort than it's worth.
> 
> If it comes down to that, I can get behind the `/v5` version being >=v1.20 minimum version

I basically just implemented it in less than an hour. Can you maybe have a look at that PR?

validator.go
oxisto: I agree, that's too many ways. Previously the Verify helpers were needed because you could not configure which claims were required / and or optional. Since all of that is now possible with parser options there is no good reason to expose the `VerifyXXX` functions anymore. I would still keep the Validate method public for now, even though its currently even impossible to directly create a `*validator` outside of the package, since we made all of the constructors private (for now).

I am also thinking about returning an error from `Verify`, but this needs some more re-write since they are currently basically returning the bool from a time/string comparison, and I probably need a helper function to turn a bool into either `nil` (true) or `ErrXXX` (false).

validator.go
oxisto: > I tried this out, and this interface helps glue together the claims in the package and any custom validation, maybe this is the least-worst option?
> 
> ```go
> type MyCustomClaims struct {
> 	Foo string `json:"foo"`
> 	jwt.RegisteredClaims
> }
> 
> var _ jwt.ClaimsValidator = (*MyCustomClaims)(nil)
> 
> func (m MyCustomClaims) Validate() error {
> 	if m.Foo != "bar" {
> 		return errors.New("must be foobar")
> 	}
> 	return nil
> }
> ```

Looks good. I implemented this in #274

validator.go
oxisto: Done, in #274

validator.go
mfridman: Sure thing. I'll take a look at that there and mark this conversation resolved here.

validator.go
mfridman: Before I jump into that PR, wdyt about unexporting the `*validator` methods? It's unexported anyways and is technically an implementation detail within this package .. might help clean up some things. And if necessary we can always export if there's a good reason.

oxisto: > Before I jump into that PR, wdyt about unexporting the `*validator` methods? It's unexported anyways and is technically an implementation detail within this package .. might help clean up some things. And if necessary we can always export if there's a good reason.

I have unexported all of them except the `Validate` method.

oxisto:
I think we are ready for a first `v5.0-rc1` or something similar, @mfridman ?

mfridman (approved):

MIGRATION_GUIDE.md
```diff
@@ -1,4 +1,82 @@
-## Migration Guide (v4.0.0)
+# Migration Guide (v5.0.0)
```
mfridman, line 1: Nicely written.

Issue #188, opened 2022-03-29
TimeFunc being static prevents from using this package in a parallel test


jsgoupil:
I tried to bring this `TimeFunc` on the `ParserOptions`, then pass it in the `Claims`
But I just saw a recent retraction of something very similar.

https://github.com/golang-jwt/jwt/pull/184

So, I would believe you will not accept a PR like mine until v5, but having the `TimeFunc()` local makes more sense for testing

https://github.com/jsgoupil/jwt/commit/a6c0c037dc35f72e1afc9a9628ae10837d4cc5bb

I hope you can consider this bug for a future release.

mfridman:
Thanks for opening this issue and linking to the PR. 

I suggest keeping it open (maybe adding a /v5 label) so we make sure to consider it if/when thinking through what a future release would look like.

oxisto:
> I tried to bring this `TimeFunc` on the `ParserOptions`, then pass it in the `Claims` But I just saw a recent retraction of something very similar.
> 
> #184
> 
> So, I would believe you will not accept a PR like mine until v5, but having the `TimeFunc()` local makes more sense for testing
> 
> [jsgoupil@a6c0c03](https://github.com/jsgoupil/jwt/commit/a6c0c037dc35f72e1afc9a9628ae10837d4cc5bb)
> 
> I hope you can consider this bug for a future release.

Did you by any chance have a look at the `v5` branch. I added a feature that hopefully solves your issue here:
https://github.com/golang-jwt/jwt/pull/234/files#diff-72f644d282bce7e136d0f762de1b9d5f3e447bfa88e5e0c4974ac2479107ddc9R40-R47

PR #236, opened 2022-08-27
New validation API

oxisto:
Some guidelines in designing the new validation API

* Previously, the `Valid` method was placed on the claim, which was always not entirely semantically correct, since the validity is concerning the token, not the claims. Although the validity of the token is based on the processing of the claims (such as `exp`). Therefore, the function `Valid` was removed from the `Claims` interface and the single canonical way to retrieve the validity of the token is to retrieve the `Valid` property of the `Token` struct.
* The previous fact was enhanced by the fact that most claims implementations had additional exported `VerifyXXX` functions, which are now removed
* All validation errors should be comparable with `errors.Is` to determine, why a particular validation has failed 
* Developers want to adjust validation options. Popular options include:
  * Leeway when processing exp, nbf, iat (#98, #237)
  * Setting the time manually, especially for testing (#188)
  * Not verifying `iat`, since this is actually just an informational claim. When purely looking at the standard, this should probably the default (#175)
  * Verifying `aud` by default, which actually the standard sort of demands. We need to see how strong we want to enforce this
* Developers want to create their own claim types, mostly by embedding one of the existing types such as `RegisteredClaims`.
  * Sometimes there is the need to further tweak the validation of a token by checking the value of a custom claim. Previously, this was possibly by overriding `Valid`. However, this was error-prone, e.g., if the original `Valid` was not called. Therefore, we should provide an easy way for *additional* checks, without by-passing the necessary validations

This leads to the following two major changes:

* The `Claims` interface now represents a set of functions that return the mandatory claims represented in a token, such as `exp`, and so on, rather than just a `Valid` function. This is also more semantically correct and makes our codebase smaller and way more maintainable.
* All validation tasks are offloaded to a new (optional) `Validator`, which can also be configured with appropriate options. If no custom validator was supplied, a default one is used.

Some smaller goodies:

* This introduces a new convenience feature for map claims: `ParseXXX` (naming to change) functions, that make it easy to parse a key as a certain JWT special type such as numeric date.

Remaining tasks

- [x] Provide a way for custom claims for additional verification
- [x] Update examples with validator goodness
- [x] Add option to skip `iat`
- [x] Moving the `TimeFunc` to the validator instead of a global function
- [x] Make `aud` verification default-ish, at least if a new validator is provided


oxisto:
An initial version of this is no ready to review. I will add some more tests once we agree on a more or less stable API.

MicahParks:
I haven't review these changes thoroughly, but glancing at the PR description, I feel neutral or favorable about the suggested changes. I'd be happy to review this PR if requested, but that would have to be next week at the earliest due to my current obligations.

I'm interested in resolving this PR for two reasons:
1. I'm concerned leaving this PR unresolved will cause a fork to emerge. I found it unfavorable with `github.com/dgrijalva/jwt-go` how many forks emerged for valid use cases or bug fixes.
2. Merging any breaking changes into this repository forces me to make an update to `github.com/MicahParks/keyfunc`. See the draft PR here: https://github.com/MicahParks/keyfunc/pull/46.

oxisto:
> 



> I haven't review these changes thoroughly, but glancing at the PR description, I feel neutral or favorable about the suggested changes. I'd be happy to review this PR if requested, but that would have to be next week at the earliest due to my current obligations.
> 
> I'm interested in resolving this PR for two reasons:
> 
> 1. I'm concerned leaving this PR unresolved will cause a fork to emerge. I found it unfavorable with `github.com/dgrijalva/jwt-go` how many forks emerged for valid use cases or bug fixes.
> 2. Merging any breaking changes into this repository forces me to make an update to `github.com/MicahParks/keyfunc`. See the draft PR here: [Experimental support for jwt v5 MicahParks/keyfunc#46](https://github.com/MicahParks/keyfunc/pull/46).

Your feedback on this PR would be most welcome!

b4k3r:
What is ETA for v5?

oxisto:
> What is ETA for v5?

There is no fixed ETA. It is kind of relying on the co-maintainers having some time to review this PR :)

oxisto:
@MicahParks did you already have time to look at this a more closely?

MicahParks:
> @MicahParks did you already have time to look at this a more closely?

Not yet, sorry. But it's still on my list of things to do.

validator.go
```diff
@@ -0,0 +1,201 @@
...
+	// verifyIat specifies whether the iat (Issued At) claim will be verified.
+	// According to https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6 this
+	// only specifies the age of the token, but no validation check is
+	// necessary. However, if wanted, it can be checked if the iat is
+	// unrealistic, i.e., in the future.
+	verifyIat bool
+
+	// expectedAud contains the audiences this token expects. Supplying an empty
+	// string will disable aud checking.
+	expectedAud string
```
MicahParks, line 27: Perhaps this should be `[]string` or the type currently named `ClaimStrings`?

validator.go
```diff
@@ -0,0 +1,201 @@
...
+	if vErr.valid() {
+		return nil
+	}
+
+	return vErr
+}
+
+// VerifyAudience compares the aud claim against cmp.
+// If required is false, this method will return true if the value matches or is unset
+func (v *Validator) VerifyAudience(claims Claims, cmp string, req bool) bool {
```
MicahParks, line 98: Perhaps the `cmp` value should be `[]string` or the type currently named `ClaimStrings`?

validator.go
```diff
@@ -0,0 +1,201 @@
...
+	verifyIat bool
+
+	// expectedAud contains the audiences this token expects. Supplying an empty
+	// string will disable aud checking.
+	expectedAud string
+}
+
+type customValidationType interface {
+	CustomValidation() error
+}
```
MicahParks, line 32: It seems to me this new interface, `customValidationType`, could be implemented by a package user to inject their own custom validation of claims. If this is the case, I think this interface should be exported and this feature should be documented, perhaps in the `README.md`.

May I suggest renaming it to `CustomValidation` and its method to `Validate`?

parser_option.go
```diff
@@ -27,3 +27,9 @@ func WithoutClaimsValidation() ParserOption {
 		p.SkipClaimsValidation = true
 	}
 }
+
+func WithValidator(v *Validator) ParserOption {
```
MicahParks, line 31:
```suggestion
// WithValidator is an option to include a claims validator.
func WithValidator(v *Validator) ParserOption {
```

validator.go
```diff
@@ -0,0 +1,201 @@
...
+	return now.After(t) || now.Equal(t)
+}
+
+func verifyIss(iss string, cmp string, required bool) bool {
+	if iss == "" {
+		return !required
+	}
+	if subtle.ConstantTimeCompare([]byte(iss), []byte(cmp)) != 0 {
+		return true
+	} else {
```
MicahParks, line 198: golint:
> validator.go:198:9: if block ends with a return statement, so drop this else and outdent its block

validator.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Apply the validator options
+	for _, o := range opts {
+		o(v)
+	}
+
+	return v
+}
+
+func (v *Validator) Validate(claims Claims) error {
```
MicahParks, line 45:
```suggestion
// Validate validates the given claims. It will also perform any custom validation if claims implements the CustomValidator interface.
func (v *Validator) Validate(claims Claims) error {
```

validator.go
```diff
@@ -0,0 +1,201 @@
...
+
+// VerifyIssuer compares the iss claim against cmp.
+// If required is false, this method will return true if the value matches or is unset
+func (v *Validator) VerifyIssuer(claims Claims, cmp string, req bool) bool {
+	return verifyIss(claims.GetIssuer(), cmp, req)
+}
+
+// ----- helpers
+
+func verifyAud(aud []string, cmp string, required bool) bool {
```
MicahParks, line 143: It seems to me this function is copied from another file. If that's the case, I know you didn't write this code, but how do you feel about making a change based on my below statement?

I don't find the `subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0` favorable over a string comparison, but perhaps I don't fully understand its purpose.

validator.go
```diff
@@ -0,0 +1,201 @@
...
+	var stringClaims string
+	for _, a := range aud {
+		if subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0 {
+			result = true
+		}
+		stringClaims = stringClaims + a
+	}
+
+	// case where "" is sent in one or many aud claims
+	if len(stringClaims) == 0 {
```
MicahParks, line 159:
```suggestion
	if stringClaims == "" {
```

validator.go
```diff
@@ -0,0 +1,201 @@
...
+
+	t := (*nbf).Add(-skew)
+	return now.After(t) || now.Equal(t)
+}
+
+func verifyIss(iss string, cmp string, required bool) bool {
+	if iss == "" {
+		return !required
+	}
+	if subtle.ConstantTimeCompare([]byte(iss), []byte(cmp)) != 0 {
```
MicahParks, line 196: Another case of code that seems to have come from another file and used `crypto/subtle`. Similar to my comment in `verifyAud`, what do you think about this case?

parser.go
```diff
@@ -82,8 +87,12 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
 
 	// Validate Claims
 	if !p.SkipClaimsValidation {
-		if err := token.Claims.Valid(); err != nil {
+		// Make sure we have at least a default validator
+		if p.validator == nil {
+			p.validator = DefaultValidator
```
MicahParks, line 92:
```suggestion
			p.validator = NewValidator()
```

parser.go
```diff
@@ -7,6 +7,9 @@ import (
 	"strings"
 )
 
+// DefaultValidator is the default validator that is used, if no custom validator is supplied in a Parser.
+var DefaultValidator = NewValidator()
```
MicahParks, line 11: Perhaps this variable could be unexported or removed? Can I have an example use case for a package write to read or write to this global variable?

MicahParks:
@oxisto Thank you for the reminder yesterday, I've reviewed the PR today and will be posting it shortly.

Apologies in advance for any comments of mine that suggest improper or inconsistent things for this code base. I haven't familiarized myself with its entirety. I'm mostly familiar with the code surrounding `jwt.Keyfunc` and what I've learned when reviewing this PR.

Also, a general comment, it seems some comments were removed on a few exported assets such as `ParseWithClaims`. I don't think that's in this PR here, but it would be good to get those back in the code base before `v5` is released.

MicahParks (changes requested):

oxisto:
> @oxisto Thank you for the reminder yesterday, I've reviewed the PR today and will be posting it shortly.
> 
> Apologies in advance for any comments of mine that suggest improper or inconsistent things for this code base. I haven't familiarized myself with its entirety. I'm mostly familiar with the code surrounding `jwt.Keyfunc` and what I've learned when reviewing this PR.
> 
> Also, it seems some comments were removed on a few exported assets such as `ParseWithClaims`. I don't think that's in this PR here, but it would be good to get those back in the code base before `v5` is released.

Thanks for the review! Yes it could be that a few comments were lost. I wanted to first get a feel of the direction of the implementation and then clean it up with further tests and documentation.

claims.go
```diff
@@ -1,177 +1,15 @@
...
-	if subtle.ConstantTimeCompare([]byte(iss), []byte(cmp)) != 0 {
-		return true
-	} else {
-		return false
-	}
+	GetExpirationTime() *NumericDate
+	GetIssuedAt() *NumericDate
+	GetNotBefore() *NumericDate
+	GetIssuer() string
+	GetAudience() ClaimStrings
```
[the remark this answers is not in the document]
oxisto: I would prefer that as well, but the problem then is that field names and function names must not be identical and the `RegisteredClaims` struct already has fields like `Issuer` and `Audience`. The only way around it would be to introduce new names that hopefully do not clash.

map_claims.go
```diff
@@ -2,150 +2,88 @@ package jwt
...
-		return !req
-	}
-
-	switch nbf := v.(type) {
-	case float64:
-		if nbf == 0 {
-			return verifyNbf(nil, cmpTime, req)
+// ParseClaimsString tries to parse a key in the map claims type as a
+// [ClaimsStrings] type, which can either be a string or an array of string.
+func (m MapClaims) ParseClaimsString(key string) ClaimStrings {
```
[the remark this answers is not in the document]
oxisto: That is true. I didn't want to change too many things but maybe since we are any way restructuring quite a bit it could be ok. Maybe we can keep a type alias for `ClaimStrings` and deprecate it and remove it a future version.

map_claims_test.go
```diff
@@ -1,10 +1,7 @@
 package jwt
 
-import (
-	"testing"
-	"time"
-)
-
+/*
+TODO(oxisto): Re-enable tests with validation API
```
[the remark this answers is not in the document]
oxisto: Yes :)

validator.go
oxisto: The way I understand it is that a system that checks a JWT for its validity must identify itself with one specific audience (e.g. its server URL), so that is why it is only a single string here. The JWT itself can be sent to multiple audiences (several servers that consume it), so that is why it is an array on the claim itself.

Does that make sense?

validator.go
oxisto: Reasoning see above

validator.go
oxisto: > It seems to me this new interface, `customValidationType`, could be implemented by a package user to inject their own custom validation of claims. If this is the case, I think this interface should be exported and this feature should be documented, perhaps in the `README.md`.

Yes that is exactly the use case! Including this in the README makes perfect sense.

> 
> May I suggest renaming it to `CustomValidation` and its method to `Validate`?

The way it works is that it checks *on top* of the checks performed by jwt and therefore it wanted something that sounds like an additional verification of custom claims, rather than something that sounds like it can override basic validation.

validator.go
oxisto: I think this is corrected on main now anyway.

validator.go
oxisto: Puh, good question. The general idea is that all things "secret" should be compared in constant time to not leak any information about them via side channels (such as time). One could argue that the audience is not really secret, but I would keep this as is for now.

validator.go
oxisto: same as above

validator.go
```diff
@@ -0,0 +1,202 @@
...
+	return now.Before((*exp).Add(+skew))
+}
+
+func verifyIat(iat *time.Time, now time.Time, required bool, skew time.Duration) bool {
+	if iat == nil {
+		return !required
+	}
+
+	t := (*iat).Add(-skew)
+	return now.After(t) || now.Equal(t)
```
sermojohn, line 181: These comparisons could be merged to `!now.Before(t)`.

validator.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func verifyNbf(nbf *time.Time, now time.Time, required bool, skew time.Duration) bool {
+	if nbf == nil {
+		return !required
+	}
+
+	t := (*nbf).Add(-skew)
+	return now.After(t) || now.Equal(t)
```
sermojohn, line 190: Similarly as [here](https://github.com/golang-jwt/jwt/pull/236/files#r1005623046).

sermojohn:
Very nice work!

This PR enables parsing and validation of the token using different APIs, so a token can be parsed once and validated by different validators.

This is very useful for implementation of token validation profiles described in the [JSON Web Token Best Current Practices RFC](https://datatracker.ietf.org/doc/html/rfc8725#section-3.12).

validator.go
```diff
@@ -0,0 +1,202 @@
...
+	// verifyIat specifies whether the iat (Issued At) claim will be verified.
+	// According to https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6 this
+	// only specifies the age of the token, but no validation check is
+	// necessary. However, if wanted, it can be checked if the iat is
+	// unrealistic, i.e., in the future.
+	verifyIat bool
+
+	// expectedAud contains the audiences this token expects. Supplying an empty
+	// string will disable aud checking.
+	expectedAud string
```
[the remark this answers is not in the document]
oxisto: As mentioned, checking the issuer is up to the application (such as an OpenID connect client), since it is only meaningful to the application. I can check whether it makes sense to include a simple string comparison, similar to the audience.

parser.go
oxisto: True, it can be removed.

validator.go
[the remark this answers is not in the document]
MicahParks: > the JWKS repo 

I assume this is referencing [`keyfunc`](https://github.com/MicahParks/keyfunc), but I also have [`jwkset`](https://github.com/MicahParks/jwkset) now :slightly_smiling_face: `jwkset` is currently in pre-release, so there's likely to be breaking changes.

Validating `iss` is not planned for `keyfunc` currently. Without refreshing myself on the project, I think it would be an outlier in the `keyfunc.Options`. If it were included, it likely increase the scope of the project to other things like `aud` and `sub` at least. If anyone feels it should be in the project, feel free to open an issue. I'll take more time to consider it.

Currently, the `jwkset` package does not have any sort of client, so any validation of `aud` would be in the distant future.

> But in the end, the application must check if the issuer corresponds to the key and as this is application specific, we probably cannot validate this in the library.

Perhaps there's something I'm missing here that would invalidate my upcoming suggestion. RFC 7519 notes for both `iss` and `aud` that they are "generally application specific". With that, I think they both have an equal argument to be included in `Valdiator` options. It seems to me that since the `Valdiator` options can be given by the package user, the "application specific" behavior can be applied through this new API, if the options are present.

* `iss`: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1
* `aud`: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3

Now that it's been mentioned, I think that _both_ `iss` and `sub` should be included in this PR as options in the same way that `aud` currently is. I know it's technically feasible for a package user to check those manually post JWT validation, but if `aud` is included in the `Validator`, I'm under the opinion that **all standardized claims** that might be checked for JWT validity should have an option, including ones that aren't mentioned in this thread. This isn't to mention the required validity checks, like `exp`.

I think it might also be a good idea to make an exported Go data structure that contains all `Validator` options like `iat`, `aud`, leeway, timeFunc, etc. This data structure would have an associated function to apply the options to the `Validator`. This would help users identify options for `Validator` without finding needing to look for all functions that return have a return type of `ValidatorOption`.

This data structure could gain more fields over time.

Something like this, perhaps:
```go
type ValidatorOptions struct {
	ExpectedAudience string
	Leeway           time.Duration
	TimeFunc         func() time.Time
	ValidateIssuedAt bool
}

func WithOptions(options ValidatorOptions) ValidatorOption {
	return func(v *Validator) {
		v.expectedAud = options.ExpectedAudience
		v.leeway = options.Leeway
		v.timeFunc = options.TimeFunc
		v.verifyIat = options.ValidateIssuedAt
	}
}
```

oxisto:
Thanks for the initial bunch of reviews, @MicahParks, @sermojohn. I incorporated most / all of your feedback. In re-enabling the map claim tests I stumbled upon a design flaw of the first iteration, in which the `GetXXX` functions of the `Claims` interface would not return an error. However, this is necessary for the `MapClaims` struct in order to differentiate between a non-existing value and an erroneous value.

validator.go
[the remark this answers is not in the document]
oxisto: > > the JWKS repo
> 
> I assume this is referencing [`keyfunc`](https://github.com/MicahParks/keyfunc), but I also have [`jwkset`](https://github.com/MicahParks/jwkset) now 🙂 `jwkset` is currently in pre-release, so there's likely to be breaking changes.

Yeah I meant `keyfunc`. :)

> 
> Validating `iss` is not planned for `keyfunc` currently. Without refreshing myself on the project, I think it would be an outlier in the `keyfunc.Options`. If it were included, it likely increase the scope of the project to other things like `aud` and `sub` at least. If anyone feels it should be in the project, feel free to open an issue. I'll take more time to consider it.
> 
> Currently, the `jwkset` package does not have any sort of client, so any validation of `aud` would be in the distant future.
> 
> > But in the end, the application must check if the issuer corresponds to the key and as this is application specific, we probably cannot validate this in the library.
> 
> Perhaps there's something I'm missing here that would invalidate my upcoming suggestion. RFC 7519 notes for both `iss` and `aud` that they are "generally application specific". With that, I think they both have an equal argument to be included in `Valdiator` options. It seems to me that since the `Valdiator` options can be given by the package user, the "application specific" behavior can be applied through this new API, if the options are present.
> 
> * `iss`: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1
> * `aud`: https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3
> 
> Now that it's been mentioned, I think that _both_ `iss` and `sub` should be included in this PR as options in the same way that `aud` currently is. I know it's technically feasible for a package user to check those manually post JWT validation, but if `aud` is included in the `Validator`, I'm under the opinion that **all standardized claims** that might be checked for JWT validity should have an option, including ones that aren't mentioned in this thread. This isn't to mention the required validity checks, like `exp`.

I can definitely include an "require IAT/ISS" similar to aud, but the issue is that the [rfc8725](https://datatracker.ietf.org/doc/html/rfc8725#section-3.8) actually suggest this keybinding validation, for example in OpenID connect by looking up the JWKS in the issuer URL; and this is something which I do not see in this library.

> 
> I think it might also be a good idea to make an exported Go data structure that contains all `Validator` options like `iat`, `aud`, leeway, timeFunc, etc. This data structure would have an associated function to apply the options to the `Validator`. This would help users identify options for `Validator` without finding needing to look for all functions that return have a return type of `ValidatorOption`.
> 
> This data structure could gain more fields over time.
> 
> Something like this, perhaps:
> 
> ```go
> type ValidatorOptions struct {
> 	ExpectedAudience string
> 	Leeway           time.Duration
> 	TimeFunc         func() time.Time
> 	ValidateIssuedAt bool
> }
> 
> func WithOptions(options ValidatorOptions) ValidatorOption {
> 	return func(v *Validator) {
> 		v.expectedAud = options.ExpectedAudience
> 		v.leeway = options.Leeway
> 		v.timeFunc = options.TimeFunc
> 		v.verifyIat = options.ValidateIssuedAt
> 	}
> }
> ```

I am a bit torn with this idea, because I think there is only a limited set of use-cases which require you to set *all* values. Instead we should rather work on a proper README / documentation of the API; something which I any way wanted to tackle: Redesign the current README and include short examples in there.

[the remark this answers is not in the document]
oxisto: There is one more fuzzy detail: The `aud`, `iss`, `sub` claims are all OPTIONAL according to the JWT RFC. However, if I as a developer use `WithAudience` / `WithIssuer` / `WithSubject` I guess I would *assume* that the claim is there (aka REQUIRED), and otherwise fail, correct? I think this behaviour should be ok from the RFC side. The question would be: does that make sense? Should we make it configurable?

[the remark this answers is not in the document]
MicahParks: > for example in OpenID connect by looking up the JWKS in the issuer URL; and this is something which I do not see in this library.

I agree that this example seems out of scope for this project.

> Instead we should rather work on a proper README / documentation of the API

This accomplishes the same goal as my suggestion, so it sounds good to me :slightly_smiling_face:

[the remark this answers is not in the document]
MicahParks: > There is one more fuzzy detail: The `aud`, `iss`, `sub` claims are all OPTIONAL according to the JWT RFC. However, if I as a developer use `WithAudience` / `WithIssuer` / `WithSubject` I guess I would _assume_ that the claim is there (aka REQUIRED), and otherwise fail, correct? I think this behaviour should be ok from the RFC side. The question would be: does that make sense? Should we make it configurable?

If I were a package user and I included `WithAudience` I would expect there to be an error if `aud` was not present. But maybe I'm poisoned with knowledge.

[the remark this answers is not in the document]
MicahParks: > Should we make it configurable?

I will suggest to make the REQUIRED presence when `WithAudience` / `WithIssuer` / `WithSubject` is set and not to let it be configurable. I think that would be best for _most use_ cases.

If a package user has a use case where it could be optional, I believe each of these claims can be checked after `Validator` is finished.

> I think that would be best for _most use_ cases.

I may be making too much of an assumption here, so anyone can contradict that.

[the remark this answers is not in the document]
oxisto: I agree 100 % @MicahParks. Then I'll implement the remaining `WithXXX`... Maybe another idea that comes into mind. While the string comparison functions are pretty much straightforward, at least for `subject` maybe it makes sense to have something like `WithSubjectPattern`; because you might want to validate the subject according to some business logic. On the other hand this could easily be done in the `CustomValidation` of your own claim. Not sure if this would be over-engineered.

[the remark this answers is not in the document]
oxisto: > > Maybe another idea that comes into mind. While the string comparison functions are pretty much straightforward, at least for `subject` maybe it makes sense to have something like `WithSubjectPattern`; because you might want to validate the subject according to some business logic. On the other hand this could easily be done in the `CustomValidation` of your own claim. Not sure if this would be over-engineered.
> 
> I, personally, would leave that out of the initial PR. If there's a good use case for it, it could be added in the future if prompted by an issue.

I'll keep it in a separate branch/PR then. I did a quick hack of it and I quite like it. Looks like this:
```
claims := RegisteredClaims{Subject: "MyUser"}
NewValidator(
  WithSubjectPattern(HasPrefix("My")),
).Validate(claims)
```

validator.go
```diff
@@ -0,0 +1,267 @@
...
+	var time *time.Time = nil
+
+	iat, err := claims.GetIssuedAt()
+	if err != nil {
+		return false
+	} else if iat != nil {
+		time = &iat.Time
+	}
+
+	return verifyIat(time, cmp, req, v.leeway)
```
sermojohn, line 162: I could find the leeway and validation suggested only for `exp` and `nbf` registered claims in the [rfc](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6), but not for `iat`.

```
This claim can be used to determine the age of the JWT.
```

Just pointing here, in case we want to avoid verification if IssueAt claim

oxisto: That's a good point. 

I think we could argue either way. In the standard, this is just an information about the age and no check of any kind is required or mentioned. On the other hand, if we want to check for plausibility of that age it might make sense to include the same leeway as for the other time-related properties.

sermojohn: It would make sense to me to make this validation optional. This would prevent any possible issue with Identity Providers (JWT issuers).

oxisto: It is. You need to specify the `WithIssuedAt` option in the validator, otherwise it will not be validated.

sermojohn: > It would make sense to me to make this validation optional.

Sorry, my comment was not correct. I meant to say the leeway should be optional, not just the validation.

oxisto: Well it is optional, if you don't specify any leeway it will check for the strict time on any of the time-based claims. Only making it "even more optional" for IAT would be overkill in my opinion.

oxisto:
Anything to add @mfridman ? I would
Love to get a v5 version out for Christmas :)

validator.go
```diff
@@ -0,0 +1,267 @@
+package jwt
+
+import (
+	"crypto/subtle"
+	"time"
+)
+
+// Validator is the core of the new Validation API. It can either be used to
```
mfridman, line 8: Something about using this API just doesn't feel idiomatic and it feels like a bit of work to get a parser with leeway support.

```
parser := jwt.NewParser(
	jwt.WithValidator(
		jwt.NewValidator(
			jwt.WithLeeway(time.Minute),
		),
	),
)
```

> .. or used standalone to validate an already parsed [Claim]

Does the Validator need to be exposed externally? What is the use case for a standalone validator?

validator_option.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// WithSubject configures the validator to require the specified subject in the
+// `sub` claim. Validation will fail if a different subject is specified in the
+// token or the `sub` claim is missing.
+//
+// NOTE: While the `sub` claim is OPTIONAL is a JWT, the handling of it is
+// application-specific. Since this validation API is helping developers in
+// writing secure application, we decided to REQUIRE the existence of the claim.
+func WithSubject(sub string) ValidatorOption {
```
mfridman, line 69: Typically a parser is created and passed down in an application, so I'm not sure how much sense there is in defining a "top-level" validator for a subject, which would be checked on a per-request basis.

The issuer and audience make sense, but this subject one only applies if the Validator is kept exported. Related to the comment above wrt Validator being exposed.

validator.go
```diff
@@ -0,0 +1,267 @@
...
+	if v.timeFunc != nil {
+		now = v.timeFunc()
+	} else {
+		now = time.Now()
+	}
+
+	// We always need to check the expiration time, but the claim itself is OPTIONAL
+	if !v.VerifyExpiresAt(claims, now, false) {
+		vErr.Inner = ErrTokenExpired
+		vErr.Errors |= ValidationErrorExpired
```
mfridman, line 74: What do you think about these operators, can we remove them in favor of more verbose but simplified logic?

claims.go
```diff
@@ -1,177 +1,16 @@
...
 
-import (
-	"crypto/subtle"
-	"fmt"
-	"time"
-)
-
-// Claims must just have a Valid method that determines
-// if the token is invalid for any supported reason
+// Claims represent any form of a JWT Claims Set according to
```
mfridman, line 3: Have you considered an alternate scenario where we remove this (and map claims) entirely?

It might be a bad idea, but just throwing this thought out .. the user still has to assert the claims.

mfridman (commented):
Great job, thanks for taking a pass at this and cleaning things up. 🎉

I'm on board with most of these changes. The only thing that's sticking out to me is the Validator being exposed to the user.

validator.go
oxisto: > Something about using this API just doesn't feel idiomatic and it feels like a bit of work to get a parser with leeway support.
> 
> ```
> parser := jwt.NewParser(
> 	jwt.WithValidator(
> 		jwt.NewValidator(
> 			jwt.WithLeeway(time.Minute),
> 		),
> 	),
> )
> ```

You are right, something like this might be easier I guess:
```golang
parser := jwt.NewParser(jwt.WithLeeway(time.Minute))
```

> 
> > .. or used standalone to validate an already parsed [Claim]
> 
> Does the Validator need to be exposed externally? What is the use case for a standalone validator?

Someone in the previous comments mentioned something about different validation profiles, so not entirely sure. I would like to keep the option though.

I will play around a little bit whether I can include validation options directly in the `NewParser`.

claims.go
oxisto: I think a lot of people still use map claims, so I am not 100 % convinced tbh.

oxisto:
> Great job, thanks for taking a pass at this and cleaning things up. 🎉
> 
> I'm on board with most of these changes. The only thing that's sticking out to me is the Validator being exposed to the user.

I made another cleanup pass and this should be fairly straightforward now. All validation options are now parser options and this has the added benefit that they also work with the `Parse` and `ParseWithClaims` functions directly:

```golang
		return []byte("AllYourBase"), nil
	}, jwt.WithLeeway(5*time.Second))
```

or as part of a new parser

```golang
parser := NewParser(jwt.WithLeeway(5*time.Second))
```

If necessary, it is still possible to create a new (standalone) validator with `NewValidator:
```golang
val := NewValidator(jwt.WithLeeway(5*time.Second))
```

This is fairly close to what @dgrijalva originally had in the "old" v4 validation API.

validator.go
mfridman: Oh yes for sure, outside the scope of this PR. But one of the things I'd love as part of the `v5` is cleaning up these errors.

I wasn't sure if there was another place we could keep a bucket/wishlist of `v5` items?

validator.go
mfridman: What do you think about moving the validator to an `internal` package, and exposing it in the public API if users request it? It's easy to expose, but once it's part of the public API then it's not possible to remove.

This is the only part I'm struggling with.

I'm a bit worried decoupling parsing/verification and validation will cause confusion.

validator.go
oxisto: Here: https://github.com/golang-jwt/jwt/pull/234

validator.go
oxisto: With the recent changes, I can just unexport the `validator` struct for now and we can see whether there is additional demand to make it public.

mfridman: Thanks for making that change 🙏

mfridman:
Overall the new API looks good from my side. I like how we're frontloading the options in one place.

The only options I'd suggest removing are `WithSubject`, `WithIssuer` and `WithAudience` .. but I also see how they can be useful. So not a blocker.

Happy to give this a stamp and merge it into the `/v5` branch. Then we can do another pass, does that seem reasonable?

claims.go
mfridman: Ye that's fair. My thought was .. there are too many ways to pass "claims"

1. map claims (default)
2. RegisteredClaims (structured, arguably could be the default?)
3. Custom claims that either implement Claims or embed RegisteredClaims

If we remove `Claims` interface entirely, and take an `any` .. would that simply the overall API?

Probably a bad idea, figured I'd bring it up but I think you're probably right .. a lot of people use map claims.

oxisto: I think it actually gets easier with the new interface, because you do not need to implement any validation on your claim, but rather you can use any kind of struct/map type that supplies the needed standard claim values, such as exp, iat, etc. I think that is a pretty straight forward way of modelling claims now.

oxisto:
> Overall the new API looks good from my side. I like how we're frontloading the options in one place.
> 
> The only options I'd suggest removing are `WithSubject`, `WithIssuer` and `WithAudience` .. but I also see how they can be useful. So not a blocker.
> 
> Happy to give this a stamp and merge it into the `/v5` branch. Then we can do another pass, does that seem reasonable?

Yap, looks good. Waiting for your official approval to merge it into `v5` ;)

mfridman (approved):
lgtm. Let's merge this into the `v5` branch and then do another few passes of reviews.

Apologies took me so long to get to this.

tschaub:
The old `claims.Valid()` method made it convenient to validate at least the `exp` claim without verifying the signature.  With the new API, is there an easy way to skip signature verification but use the other claim validation logic?

PR #246, opened 2022-10-13
fix: VerifyExpiresAt

matteosilv:
VerifyExpiresAt does not do what is expected from its description

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -244,7 +244,7 @@ func verifyExp(exp *time.Time, now time.Time, required bool) bool {
 	if exp == nil {
 		return !required
 	}
-	return now.Before(*exp)
+	return now.Before(*exp) || now.Equal(*exp)
 }
 
 func verifyIat(iat *time.Time, now time.Time, required bool) bool {
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -104,19 +104,19 @@ func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
 	mapClaims := MapClaims{
 		"exp": float64(exp),
 	}
-	want := false
+	want := true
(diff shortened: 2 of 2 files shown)


oxisto:
I am not quite sure if I understand you correctly. The description says `VerifyExpiresAt compares the exp claim against cmp (cmp < exp).` Which we do by only using `Before`.

matteosilv:
> I am not quite sure if I understand you correctly. The description says `VerifyExpiresAt compares the exp claim against cmp (cmp < exp).` Which we do by only using `Before`.

It isn't correct documentation states <=

https://github.com/golang-jwt/jwt/blob/main/map_claims.go#L35

oxisto:
> > I am not quite sure if I understand you correctly. The description says `VerifyExpiresAt compares the exp claim against cmp (cmp < exp).` Which we do by only using `Before`.
> 
> It isn't correct documentation states <=
> 
> https://github.com/golang-jwt/jwt/blob/main/map_claims.go#L35

Ah, it seams the documentation for MapClaims says <= and for StandardClaims it says <.

From a logical point of view I would say that < is correct because at the time of exp it is already expired. The additional leeway we can grant is another story. I will double check with the RFC later.

matteosilv:
VerifyExpiresAt should verify that at the provided time the token is going to be expired. Is this assumption right?

oxisto:
> VerifyExpiresAt should verify that at the provided time the token is going to be expired. Is this assumption right?

No. It checks whether the token is *not yet expired*. I agree that we should probably make the documentation of these functions clearer.

matteosilv:
In this case, I agree that strict < is correct. The name of the method and its description however are not clear enough imho

oxisto:
Superseded by #234

PR #248, opened 2022-10-15
Improve precision on NumericDate.UnmarshalJSON

arukuka:
I faced an precision issue when calling `UnmarshalJSON` for `NumericDate`.  
For example, when original unix time was `9223371974719179007`,  
`NumericDate.Unix()` shows `9223371974719178752` after `json.Unmarshal`.

```go
	var t jwt.NumericDate
	json.Unmarshal([]byte("9223371974719179007"), &t)
	fmt.Println(t.Unix()) // 9223371974719178752
```

I think that this issue is caused because code uses `float64`.

https://github.com/golang-jwt/jwt/blob/0c4e3879854669acd15ea435f2c8aada6c73810a/types.go#L88-L93

This patch use rational type of `math/big` standard package when parcing JSON to improve precision.

This patch includes test by 10bc0152ec873c8c510a6c6db7c14a0470a6c43e .  
We can check that added test is going to be failed by checkout 10bc0152ec873c8c510a6c6db7c14a0470a6c43e .

### DIFF
--- a/types.go
+++ b/types.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"math/big"
 	"reflect"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


oxisto:
Thanks for catching that. We had similar issues in the past and I though we fixed this already. I am wondering a little bit whether this has any performance drawbacks? 

My worry is a little bit that this might impact performance and in the end the issue isn't "that big" in the sense that by nature the expiration date does not need to be super precise down to the nanoseconds. In general, people are even looking towards some leeway.

I can understand that we want to get this fixed when marshalling/unmarshalling tokens, so that the values are the same.

oxisto:
https://github.com/golang-jwt/jwt/pull/200/files  is the PR I am referring to. Seems that this was only fixing the marshaling side of things.

arukuka:
I am very sorry for the very late reaction. :bow:
Thank you for your review.
If I have the next opportunity, I will create another PR based on the review comments I received.
I will close this PR.

Issue #223, opened 2022-07-15
Token without ExpiresAt runs panic


82andre:
Token without ExpiresAt runs panic, the panic has a recovery, but it would be better have an error.

Sample code: [Code in Playground ](https://go.dev/play/p/pQaxs7bhYJW)
```
package main

import (
	"bytes"
	"errors"
	"fmt"
	"io"
	"log"
	"net/http"
	"net/http/httptest"
	"strings"
	"time"

	"github.com/golang-jwt/jwt/v4"
)

func main() {
	fmt.Println("RUN createToken")
	token, err := createToken()
	fatal(err)
	runTest(token) //this RUN OK

	fmt.Println("RUN createTokenThatNotExpire")
	tokenNotExpire, err := createTokenThatNotExpire()
	fatal(err)
	runTest(tokenNotExpire) //this call panic at parser.go:85
	fmt.Println("finish")
}
func runTest(token string) {
	req := httptest.NewRequest("GET", "http://localhost/test", nil)
	req.Header.Set("Authorization", fmt.Sprintf("Bearer %v", token))
	w := httptest.NewRecorder()
	err := auth(w, req)
	fatal(err)
	res := w.Result()
	// Read the response body
	buf := new(bytes.Buffer)
	io.Copy(buf, res.Body)
	res.Body.Close()
	fmt.Println(buf.String())
	fmt.Println(token)
	fmt.Println("end.")
}

func createToken() (string, error) {
	t := jwt.New(jwt.GetSigningMethod("HS256"))
	t.Claims = &tokenClaims{
		&jwt.RegisteredClaims{
			ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Hour * 1)),
		},
		tokenInfo{"test", "1.0", "hi@hi.com.br"},
	}
	return t.SignedString(mySigningKey)
}
func createTokenThatNotExpire() (string, error) {
	t := jwt.New(jwt.GetSigningMethod("HS256"))
	t.Claims = &tokenClaims{
		&jwt.RegisteredClaims{
			//ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Hour * 1)),
		},
		tokenInfo{"test", "1.0", "hi@hi.com.br"},
	}
	return t.SignedString(mySigningKey)
}

func auth(w http.ResponseWriter, r *http.Request) error {
	tokenStr, ok := r.Header["Authorization"]
	if !ok || len(tokenStr) != 1 {
[...]

PR #255, opened 2022-11-08
No pointer embedding in the example

oxisto:
Embedding a pointer version of `jwt.RegisteredClaims` in a custom claim and passing it to `ParseWithClaims` without memory allocation will lead to a crash if an empty token is presented (see #223). Unfortunately, we cannot really check for this in the function without MAJOR reflection magic, so I fear, all we can do at the moment is adjust the example and add a better description.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -42,10 +42,13 @@ func (p *Parser) Parse(tokenString string, keyFunc Keyfunc) (*Token, error) {
 	return p.ParseWithClaims(tokenString, MapClaims{}, keyFunc)
 }
 
-// ParseWithClaims parses, validates, and verifies like Parse, but supplies a default
-// object implementing the Claims interface. This provides default values which
-// can be overridden and allows a caller to use their own type, rather than
-// the default MapClaims implementation of Claims.
+// ParseWithClaims parses, validates, and verifies like Parse, but supplies a default object implementing the Claims
+// interface. This provides default values which can be overridden and allows a caller to use their own type, rather
+// than the default MapClaims implementation of Claims.
+//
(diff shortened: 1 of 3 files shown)


parser.go
```diff
@@ -42,10 +42,13 @@ func (p *Parser) Parse(tokenString string, keyFunc Keyfunc) (*Token, error) {
...
-// object implementing the Claims interface. This provides default values which
-// can be overridden and allows a caller to use their own type, rather than
-// the default MapClaims implementation of Claims.
+// ParseWithClaims parses, validates, and verifies like Parse, but supplies a default object implementing the Claims
+// interface. This provides default values which can be overridden and allows a caller to use their own type, rather
+// than the default MapClaims implementation of Claims.
+//
+// Note: If you provide a custom claim implementation that embeds one of the standard claims (such as RegisteredClaims),
+// make sure that a) you either embed a non-pointer version of the claims or b) if you are using a pointer, allocate the
+// proper memory for it before passing in the overall claims, otherwise you might run into a panic.
```
mfridman, line 51: This seems like a misuse of the language, no? I.e., the same could be said about adding to an uninitialized map in Go, that's how the language behaves and every package out there doesn't need to call it out.

mfridman (approved):
Okay, I scanned through the GH issues and a few folks have run into this. So maybe this is for the best.

I don't think we should do any more than this.

parser.go
oxisto: Yes it is, but since people seem to be running into this issue and this is a security-related package, it makes sense to point this out explicitly.

PR #257, opened 2022-11-14
Fix the inconsistency of errors thrown

qeq66:


map_claims.go
```diff
@@ -127,19 +127,19 @@ func (m MapClaims) Valid() error {
 
 	if !m.VerifyExpiresAt(now, false) {
 		// TODO(oxisto): this should be replaced with ErrTokenExpired
```
oxisto, line 129: Can you remove the TODO as well? This changes the actual error message a *little* bit (uppercase vs. lowercase) but I like it to have more coherent so I think we can give this a go. And the uppercase error was actually a bad style anyway.

PR #259, opened 2022-12-02
Allow strict base64 decoding

AlexanderYastrebov:
By default base64 decoder works in non-strict mode which allows tweaking signatures having padding without failing validation.

This creates a potential problem if application treats token value as an identifier.

For example ES256 signature has length of 64 bytes and two padding symbols (stripped by default). Therefore its base64-encoded value can only end with A, Q, g and w. In non-strict mode last symbol could be tweaked resulting in 16 distinct token values having the same signature and passing validation.

This change adds backward-compatible global config variable DecodeStrict (similar to existing DecodePaddingAllowed) that enables strict base64 decoder mode.

See also https://github.com/golang/go/issues/15656.

Signed-off-by: Alexander Yastrebov <yastrebov.alex@gmail.com>

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -489,6 +489,7 @@ var setPaddingTestData = []struct {
 	tokenString   string
 	claims        jwt.Claims
 	paddedDecode  bool
+	strictDecode  bool
 	signingMethod jwt.SigningMethod
 	keyfunc       jwt.Keyfunc
 	valid         bool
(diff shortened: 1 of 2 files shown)


oxisto (approved):

AlexanderYastrebov:
I made it backward-compatible but IMO this could be true by default. Let me know what you think, I'll update the PR if needed.

oxisto:
> I made it backward-compatible but IMO this could be true by default. Let me know what you think, I'll update the PR if needed.

I would keep it as false by default for this version and we can think about making it true for v5.

PR #261, opened 2022-12-05
Fix the inconsistency of errors thrown

qeq66:


qeq66:
@oxisto

map_claims.go
```diff
@@ -126,20 +126,17 @@ func (m MapClaims) Valid() error {
 	now := TimeFunc().Unix()
```
oxisto, line 126: You need to remove the unused import to `errors` as well. You do not need to open/close a new PR, you can add new commits to your existing branch.

qeq66: All right, handled

oxisto (approved):
Because this slightly changes the error descriptions (lower vs. uppercase) I propose merging this into the `v5` branch.

PR #267, opened 2023-01-31
use errors that can be referenced

dmisol:
Use error values from errors.go in map_claims.go, so they can be compared.

### DIFF
--- a/map_claims.go
+++ b/map_claims.go
@@ -2,7 +2,6 @@ package jwt
 
 import (
 	"encoding/json"
-	"errors"
 	"time"
 	// "fmt"
 )
@@ -127,19 +126,19 @@ func (m MapClaims) Valid() error {
 
 	if !m.VerifyExpiresAt(now, false) {
 		// TODO(oxisto): this should be replaced with ErrTokenExpired
-		vErr.Inner = errors.New("Token is expired")
+		vErr.Inner = ErrTokenExpired
 		vErr.Errors |= ValidationErrorExpired
 	}
 
 	if !m.VerifyIssuedAt(now, false) {
 		// TODO(oxisto): this should be replaced with ErrTokenUsedBeforeIssued
-		vErr.Inner = errors.New("Token used before issued")
+		vErr.Inner = ErrTokenUsedBeforeIssued
 		vErr.Errors |= ValidationErrorIssuedAt
 	}
 
 	if !m.VerifyNotBefore(now, false) {
 		// TODO(oxisto): this should be replaced with ErrTokenNotValidYet
-		vErr.Inner = errors.New("Token is not valid yet")
+		vErr.Inner = ErrTokenNotValidYet
 		vErr.Errors |= ValidationErrorNotValidYet
 	}
 


oxisto (approved):

oxisto:
> Use error values from errors.go in map_claims.go, so they can be compared.

Thanks! We sort of tried to avoid fixing this in `v4` because it slightly could break things when people compare the actual string. However, as long as you use `errors.Is´ (which you should), things should still work.

However, we anyway refactor all of this in `v5`.  What's you opinion on this @mfridman ?

mfridman:
> > Use error values from errors.go in map_claims.go, so they can be compared.
> 
> Thanks! We sort of tried to avoid fixing this in `v4` because it slightly could break things when people compare the actual string. However, as long as you use `errors.Is´ (which you should), things should still work.
> 
> However, we anyway refactor all of this in `v5`. What's you opinion on this @mfridman ?

I think we should do this only in the `v5` release and break as little as possible in the `v4`. There's probably a lot of projects depending on this package, so trying to stay backwards-compatible is probably the most sane thing for everyone.

oxisto:
> > > Use error values from errors.go in map_claims.go, so they can be compared.
> > 
> > 
> > Thanks! We sort of tried to avoid fixing this in `v4` because it slightly could break things when people compare the actual string. However, as long as you use `errors.Is´ (which you should), things should still work.
> > However, we anyway refactor all of this in `v5`. What's you opinion on this @mfridman ?
> 
> I think we should do this only in the `v5` release and break as little as possible in the `v4`. There's probably a lot of projects depending on this package, so trying to stay backwards-compatible is probably the most sane thing for everyone (at least within a major release).
> 
> EDIT: i know it seems trivial, but it's inevitable someone is doing some string matching somewhere. And if folks want well-defined errors they should consider upgrading.

Sounds good to me, but then we should probably move our progress towards v5 a little bit ;)

mfridman:
> > > > Use error values from errors.go in map_claims.go, so they can be compared.
> > > 
> > > 
> > > Thanks! We sort of tried to avoid fixing this in `v4` because it slightly could break things when people compare the actual string. However, as long as you use `errors.Is´ (which you should), things should still work. However, we anyway refactor all of this in `v5`. What's you opinion on this @mfridman ?
> > 
> > 
> > I think we should do this only in the `v5` release and break as little as possible in the `v4`. There's probably a lot of projects depending on this package, so trying to stay backwards-compatible is probably the most sane thing for everyone (at least within a major release).
> > EDIT: i know it seems trivial, but it's inevitable someone is doing some string matching somewhere. And if folks want well-defined errors they should consider upgrading.
> 
> Sounds good to me, but then we should probably move our progress towards v5 a little bit ;)

Ye you're right @oxisto, I have a long weekend coming up. So I'll try to get as much of the reviews as possible, and then we can release a `v5.0.0-rc.1` version? Or did you want to go straight into a stable `v5.0.0` ?

Are you on [Gophers Slack](https://join.slack.com/t/gophers/shared_invite/zt-1pjl6cv77-y5mb0MKgpSkZFKukD3rzkg) by any chance?

oxisto:
> > > > > Use error values from errors.go in map_claims.go, so they can be compared.
> > > > 
> > > > 
> > > > Thanks! We sort of tried to avoid fixing this in `v4` because it slightly could break things when people compare the actual string. However, as long as you use `errors.Is´ (which you should), things should still work. However, we anyway refactor all of this in `v5`. What's you opinion on this @mfridman ?
> > > 
> > > 
> > > I think we should do this only in the `v5` release and break as little as possible in the `v4`. There's probably a lot of projects depending on this package, so trying to stay backwards-compatible is probably the most sane thing for everyone (at least within a major release).
> > > EDIT: i know it seems trivial, but it's inevitable someone is doing some string matching somewhere. And if folks want well-defined errors they should consider upgrading.
> > 
> > 
> > Sounds good to me, but then we should probably move our progress towards v5 a little bit ;)
> 
> Ye you're right @oxisto, I have a long weekend coming up. So I'll try to get as much of the reviews as possible (again, sorry for blocking you .. I just can't catch a break atm), and then we can release a `v5.0.0-rc.1` version? Or did you want to go straight into a stable `v5.0.0` ?

A first beta or release version would be a good start I think

> 
> Are you on [Gophers Slack](https://join.slack.com/t/gophers/shared_invite/zt-1pjl6cv77-y5mb0MKgpSkZFKukD3rzkg) by any chance?

Not really unfortunatly, I mainly use teams in my "professional" life and discord for everything else.

oxisto:
Since we are only aiming at this for v5 we can close this here because it is already fixed / addressed in the `v5` branch.

PR #268, opened 2023-02-09
Adding more coverage

oxisto:
This PR adds some more test coverage to the core functionalities. Not sure what % we want to achieve?

### DIFF
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -134,3 +134,56 @@ func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
 		t.Fatalf("Failed to verify claims, wanted: %v got %v", want, (got == nil))
 	}
 }
+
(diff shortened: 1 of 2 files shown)


map_claims_test.go
```diff
@@ -134,3 +134,56 @@ func TestMapClaimsVerifyExpiresAtExpire(t *testing.T) {
...
+	}
+	tests := []struct {
+		name    string
+		m       MapClaims
+		args    args
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "missing key",
```
mfridman, line 150: nit, i wish we established a pattern where we named these "missing_key" or "missing-key" so the names are copy/pasteable.

that way it's easy to do -run=TestMapClaims_ParseString/{name} without having to deal with regex or manually concatenate the name.

(this isn't the best example, but still).

Otherwise thanks for adding more tests!

PR #271, opened 2023-02-17
Added support for type parameters in the `ParseXXX` functions

oxisto:
Following the discussion in https://github.com/golang-jwt/jwt/discussions/269, this PR adds support for type parameters (or generics) in the `ParseXXX` functions, basically allowing one to directly access a specific claims type without any uncomfortable type assertions in either the `keyfunc` or the resulting `Token` struct.


oxisto:
Will continue discussion in #272

PR #274, opened 2023-02-19
More consistent way of handling validation errors

oxisto:
This PR removes the mess of the old `ValidationError` struct and replaces it with either multiple wrapped errors (in case of Go 1.20) or an implementation that emulates that behaviour using internal structs (on Go < 1.20).

It also consistently aborts the validation process now if signature validation fails, which otherwise might lead to unexpected error situations.


validator.go
```diff
@@ -69,56 +72,49 @@
...
 		}
 	}
 
-	if vErr.valid() {
+	if len(errs) == 0 {
 		return nil
 	}
 
-	return vErr
+	return errors.Join(errs...)
```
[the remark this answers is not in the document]
oxisto: That's a lie ;)

none.go
```diff
@@ -13,7 +15,7 @@ type unsafeNoneMagicConstant string
 
 func init() {
 	SigningMethodNone = &signingMethodNone{}
-	NoneSignatureTypeDisallowedError = NewValidationError("'none' signature type is not allowed", ValidationErrorSignatureInvalid)
+	NoneSignatureTypeDisallowedError = fmt.Errorf("%w: 'none' signature type is not allowed", ErrTokenUnverifiable)
```
oxisto, line 18: Note to self: should be replaced with `newError`

none.go
```diff
@@ -33,10 +35,7 @@ func (m *signingMethodNone) Verify(signingString, signature string, key interfac
 	}
 	// If signing method is none, signature must be an empty string
 	if signature != "" {
-		return NewValidationError(
-			"'none' signing method with non-empty signature",
-			ValidationErrorSignatureInvalid,
-		)
+		return fmt.Errorf("%w: 'none' signing method with non-empty signature", ErrTokenUnverifiable)
```
oxisto, line 38: Note to self: should be replaced with `newError`

parser.go
```diff
@@ -65,25 +65,25 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
...
-		}
-		return token, &ValidationError{Inner: err, Errors: ValidationErrorUnverifiable}
+		return token, newError("error while executing keyfunc", ErrTokenUnverifiable, err)
 	}
 
-	vErr := &ValidationError{}
+	// Perform signature validation
+	token.Signature = parts[2]
+	if err = token.Method.Verify(strings.Join(parts[0:2], "."), token.Signature, key); err != nil {
+		return token, newError("could not verify", ErrTokenSignatureInvalid, err)
```
oxisto, line 85: detail message is not needed here actually

parser_test.go
```diff
@@ -359,7 +334,7 @@ func TestParser_Parse(t *testing.T) {
 
 			// Parse the token
 			var token *jwt.Token
-			var ve *jwt.ValidationError
+			//var ve *jwt.ValidationError
```
oxisto, line 337: Remove

errors.go
```diff
@@ -2,120 +2,48 @@ package jwt
...
-
-// Unwrap gives errors.Is and errors.As access to the inner error.
-func (e *ValidationError) Unwrap() error {
-	return e.Inner
-}
 
-// No errors
-func (e *ValidationError) valid() bool {
-	return e.Errors == 0
+	return strings.Join(msg, ", ")
```
mfridman, line 40:
```suggestion
	return strings.Join(msg, ": ")
```

Would it be more idiomatic if these were separated by colons (`:`) ?

No strong preference, simply an observation that colon is more widely used (at least based on my experience and those who adopted pkg/errors).

mfridman (commented):
This is fantastic work! 

The new error handling reads a lot better in my opinion and hopefully, in a few releases, we can drop the older error branch and build tags entirely.

errors_go1_20.go
```diff
@@ -0,0 +1,47 @@
...
+// include more than one %w formatting directive in [fmt.Errorf].
+//
+// For example,
+//
+//	newError("no keyfunc was provided", ErrTokenUnverifiable)
+//
+// will produce the error string
+//
+//	"token is unverifiable: no keyfunc was provided"
+func newError(message string, err error, more ...error) error {
```
mfridman, line 29: Based on the comment above, we're not using `errors.Join` so we can have control over the formatting of the message?

go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/golang-jwt/jwt/v5
 
-go 1.16
+go 1.20
```
mfridman, line 3: Does this need to change to 1.20, or can this remain on the minimum version we currently support?

errors_go1_20.go
mfridman, line 29: I was a bit confused at first why this wasn't `newError(errs ...error)`, but after some more reading I quite like the explicit way we construct it, nicely done!

mfridman (approved):

errors.go
oxisto: Errors that are caused by another are concatenated using `:`, errors that are sort of "in parallel" are separated by `,` So that basically you end up with something that reads: 
```
token has invalid claims: token is expired, token is not valid yet
```

If we choose to only use `:` you would end up with 

```
token has invalid claims: token is expired: token is not valid yet
```

which would be weird, since the "not valid yet" is not a cause of "expired", but basically exists next to each other.

The new Go 1.20 `errors.Join` uses a newline to join them; we can also simulate that but this is a bit weird I think personally: 
```
token has invalid claims: token is expired
token not valid yet
```

go.mod
oxisto: I thought I reverted this back to Go 1.18 at least, I ll double check, maybe I only edited the CI file. I think per our 2-release-support strategy we can bump this to 1.18.

errors_go1_20.go
oxisto: Yes. see comment above

errors.go
mfridman: Okay, that's fair. I agree the newline is an odd choice and the inlined version (as you have it) is better.

The newline assumes it's always being printed out, but that's unlikely to be the case. Most often I wager these will end up in some structured logger and the newline will make it harder to read.

agufagit:
How do you know a jwt is expired with this approach?  In v4 you can check if an error is jwt.ValidationErrorExpired, now the struct joinedError is private, do we need to scan for phrase "token is expired" (ErrTokenExpired) in err.Error()?

this check is needed to know exactly that an Oauth2 access token is expired, and not invalidated for other reasons

PR #275, opened 2023-02-21
Adding canonical `Keyfunc` functions for RSA, ECDSA, EdDSA and HMAC

oxisto:
This PR adds ready-to-use keyfunc functions for the various signing methods. This should simplify a lot of standard use-cases and also includes a proper signing method check.

This allows for a much cleaner experience in probably 90% of all use cases:

```go
token, err := jwt.ParseWithClaims(
  tokenString, &MyCustomClaims{},
  jwt.PresharedKey([]byte("AllYourBase")),
  jwt.WithLeeway(5*time.Second),
)
```

```go
token, err := jwt.ParseWithClaims(
  tokenString, &MyCustomClaims{},
  jwt.RSAPublicKey(myKey),
  jwt.WithAudience("http://example.com"),
)
```

### DIFF
--- a/ecdsa.go
+++ b/ecdsa.go
@@ -132,3 +132,10 @@ func (m *SigningMethodECDSA) Sign(signingString string, key interface{}) ([]byte
 		return nil, err
 	}
 }
+
+// ECDSAPublicKey represents a [Keyfunc] that returns the ECDSA key specified in
+// key. Furthermore, it checks, whether the signing method matches
+// [SigningMethodECDSA].
+func ECDSAPublicKey(key *ecdsa.PublicKey) Keyfunc {
+	return secureKeyFunc(key, []string{"ES256", "ES384", "ES512"})
+}
--- a/ed25519.go
+++ b/ed25519.go
@@ -78,3 +78,10 @@ func (m *SigningMethodEd25519) Sign(signingString string, key interface{}) ([]by
 
 	return sig, nil
 }
+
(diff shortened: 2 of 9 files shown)


mfridman (approved):
I've gone back-forth on this PR. On one hand, it's a nice quality of life improvement for the most common use cases.

On the other hand, the caller needs to decide whether to use this helper function or pass their own function, using the exported structs and calling `.Alg()`.

Also, I've often needed access to the header bits for additional validation, refresh, etc. and so always constructed this myself.

Not against this, just want to make sure we're expanding the API surface area where it makes sense.

oxisto:
> I've gone back-forth on this PR. On one hand, it's a nice quality of life improvement for the most common use cases.
> 
> On the other hand, the caller needs to decide whether to use this helper function or pass their own function, using the exported structs and calling `.Alg()`.
> 
> Also, I've often needed access to the header bits for additional validation, refresh, etc. and so always constructed this myself.
> 
> Not against this, just want to make sure we're expanding the API surface area where it makes sense.

Yeah same here, the main reason for this was because we had a lot of issues that were not sure how to pass public/private keys properly. This is now somewhat mitigated by the documentation page. I still think it's a nice, clean way of using this library, though.

Maybe we can keep it as "approved", but not merge it yet and aim for a 5.1 release.

oxisto:
Closing this in favour of external libraries, such as https://github.com/MicahParks/keyfunc

Issue #168, opened 2022-02-17
Supporting alternative base64/json encoder/decoder


mattrobenolt:
Would you be open to supporting alternative base64 implementations? The one we've been using is https://github.com/segmentio/asm/tree/main/base64

It's significantly faster than the stdlib encoding, and takes shortcuts that I don't see why wouldn't be acceptable here.

Ideally, if the token parsers were abstracted to accept a `Encoding` interface, we could pass in our own, with default behaviors to maintain `encoding/base64`.

In similar vein, similar can be said for JSON codecs. The stdlib is known for having sub-optimal performance here. Swapping `encoding/json` for something like, `fastjson` (https://github.com/valyala/fastjson) would be nice.

I'm new to this project and haven't dug in much, so I don't know all the consequences of this or how to gauge it yet, but I know for us internally, we've adopted these codec alternatives for packages we control.

oxisto:
In theory, that could be an interesting option to explore. However, because this library is used by many users, we basically have two major design principles maybe should write them down somewhere :):

- Zero (0, nil, null) external dependencies
- Strict adherence to a non-API breaking way of adding features (within the same major version)

So, if we find a way that is non-breaking (and defaults for the stdlib), has no dependencies to the mentioned project (which should be doable, if we introduce interfaces here), I would not strongly object it. Personally, I do not see a major use case through. Not having any hard numbers here, but my *guess* is that the most expensive operation is the crypto part, so not sure how large the performance gain really is. I do get the idea of having the other libraries in your project and having this consistent.

Would you mind drafting up a PR?

mattrobenolt:
I can look through it. The nice part about working through interfaces here is we wouldn't need to add any external dependencies.

We can just define a `Base64Encoder` interface and `JSONEncoder` interface, and as long as libraries conformed to the expected methods, we'd allow them to be plugged in without strictly depending on them here.

I agree that crypto bits would likely outweigh this, but if you're operating on the order of tens of thousands of rps, these wins with encoding/decoding are likely a measurable %age gain.

I'll probably take a pass and see what a theoretical implementation would look like and throw together a benchmark to prove/disprove that this would be worthwhile.

oxisto:
[...]

PR #278, opened 2023-02-21
Moving `DecodeSegement` to `Parser`

oxisto:
This PR moves `DecodeSegment` to `Parser` and `EncodeSegment` to `Token`. This will allow us to configure the behaviour of these two based on options supplied on the parser or the token (creation). This removes two previously global variables and moves them to parser options `WithStrictDecoding` and `WithPaddingAllowed`.

Note, that in order to do that, I had to redesign the way signing methods work. Previously they were given a base64 encoded signature in `Verify` and were expected to return a base64 encoded version of the signature in `Sign`. However, this made it necessary to have `DecodeSegment` and `EncodeSegment` global and was a bad design because we were repeating encoding/decoding steps for *all* signing methods. Now, `Sign` and `Verify` take an already decoded signature as a `[]byte`, which feels more natural for a cryptographic operation anyway.

In addition to that, I also changed the `Signature` field on `Token` from a `string` to `[]byte` and this is also now populated with the decoded form. This is also more consistent, because the other parts of the JWT, mainly `Header` and `Claims` were already stored in decoded form in `Token`, only the signature was stored in encoded form, which was redundant with the information in `Raw`.

This also introduces a new (unexported) `tokenOption` struct which has been added to `jwt.New` and `jwt.NewWithClaims` as a varargs argument. This will allow us in the future to supply options to these functions, if we ever need to support a scenario like #168.

### DIFF
--- a/ecdsa_test.go
+++ b/ecdsa_test.go
@@ -3,6 +3,7 @@ package jwt_test
 import (
 	"crypto/ecdsa"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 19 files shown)


ecdsa_test.go
```diff
@@ -155,10 +157,24 @@ func BenchmarkECDSASigning(b *testing.B) {
 				if err != nil {
 					b.Fatalf("[%v] Error signing token: %v", data.name, err)
 				}
-				if sig == parts[2] {
+				if reflect.DeepEqual(sig, decodeSegment(b, parts[2])) {
```
mfridman, line 160: This feels more like a test, do we need to assert here?

mfridman (approved):

mfridman:
It'd be nice, eventually, if we could document the happy path (read default behaviour) without the functional options.

 And then outline which functional options are used and when.

ecdsa_test.go
oxisto: I am not quite sure what you mean

oxisto:
> It'd be nice, eventually, if we could document the happy path (read default behaviour) without the functional options.
> 
> And then outline which functional options are used and when.

https://golang-jwt.github.io/jwt/usage/parse/#with-options

ecdsa_test.go
oxisto: Ah, because it’s a benchmark you mean? Good point

PR #279, opened 2023-02-22
Add golangci-lint

mfridman:
This PR adds golangci-lint to run in CI on every merge to `main` or every PR.

It probably makes sense to have at least the [default linters](https://golangci-lint.run/usage/linters/#enabled-by-default) enabled.

More of an experiment to see if we'd like this, or not.

ps. There is a strict formatter I like, [gofumt](https://github.com/mvdan/gofumpt), but that may be too opinionated.


oxisto:
We should then remove the `reviewdog/action-staticcheck@v1` from `build`, otherwise we have two linters running. I would also upgrade the CodeQL action while we are at it.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: "1.20.x"
+          check-latest: true
+          cache: true
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.51.2
```
oxisto, line 24: Can we just set this to "latest"? One last version to update.

Makefile
```diff
@@ -0,0 +1,14 @@
+GO_TEST_FLAGS ?= -race -count=1 -v -timeout=10m
```
oxisto, line 1: Do you want to use the Makefile in the build step? Personally, I am not a big fan of Makefiles, seems to be a very "old" technology ;) Building this Go package should be straight forward.

mfridman: I think they are convenient because `make` is installed on every computer and CI. And it's a consistent way to run something like `make test`. Not using it for building per se, but more for consistently running a few go commands.

But indifferent to them.

oxisto (approved):

PR #297, opened 2023-03-30
added support to dynamic sign the claims

### LINKED ISSUE #294: Be able to skip b64 encoding on payload
Hey,
I did have a issue where I needed to prevent signing the jwt payload with b64. Is there a option to skip b64 encoding or is it possible to add it?

emilpriver:
Hello,

Added support to only sign claims in JWT if b64 header is present and has value `false`. 

This is useful if you need to create a detached JWS.

Instead of `SigningString`returning `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2YWx1ZSI6ImhlbGxvIHdvcmxkIn0` do SigningString now return `eyJhbGciOiJIUzI1NiIsImI2NCI6ZmFsc2UsInR5cCI6IkpXVCJ9.{"value":"hello world"}` Which later on is signed used `SignedString`

Also this PR adds the support b64 header should have: https://www.rfc-editor.org/rfc/rfc7797#page-4

Fixes #294

### DIFF
--- a/token.go
+++ b/token.go
@@ -71,7 +71,7 @@ func (t *Token) SigningString() (string, error) {
 		return "", err
 	}
 
-	// If payload have header b64 set as true, then don't base64 encode claims
+	// If payload have header b64 set as false, then don't base64 encode claims
 	if enabled, ok := t.Header["b64"].(bool); ok && !enabled {
 		return t.EncodeSegment(h) + "." + string(c), nil
 	}
--- a/token.go
+++ b/token.go
@@ -71,6 +71,11 @@ func (t *Token) SigningString() (string, error) {
 		return "", err
 	}
 
+	// If payload have header b64 set as true, then don't base64 encode claims
+	if enabled, ok := t.Header["b64"].(bool); ok && !enabled {
+		return t.EncodeSegment(h) + "." + string(c), nil
+	}
+
 	return t.EncodeSegment(h) + "." + t.EncodeSegment(c), nil
 }
 
--- a/token_test.go
+++ b/token_test.go
@@ -1,6 +1,8 @@
 package jwt_test
 
 import (
+	"encoding/json"
+	"strings"
 	"testing"
 
 	"github.com/golang-jwt/jwt/v5"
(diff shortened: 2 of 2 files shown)


emilpriver:
I wasn't able to assign reviewrs @dgrijalva so i'm pinging you instead.

oxisto:
I do have some concerns with this
a) We are a library aiming at JSON Web Tokens (JWTs), defined in RFC 7519. While this means that we need to implement (partially) implement JWS, we restrict our selves to the core set needed to fulfil the necessary requirements for creating the JWT. We do not aim at full support for JWS, and adding extra features for JWS is something we need to be careful about (in the sense of maintaining it)

b) RFC 7797 explicitly forbids the use of unencoded JWS payloads (the `b64 = false` option) for JWTs. It says so in the abstract and in https://www.rfc-editor.org/rfc/rfc7797#section-7. To quote: 

```
This specification updates [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519) by stating that JSON Web Tokens (JWTs) MUST NOT use the unencoded payload option defined by this specification.
```

and

```
For interoperability reasons, JSON Web Tokens [[JWT](https://www.rfc-editor.org/rfc/rfc7797#ref-JWT)] MUST NOT use "b64" with a "false" value.
```

Since this updates RFC 7519, this basically would mean that we are non-compliant to our core RFC when adding this feature. My recommendation would be to use a library that aims for full JWS support, such as https://github.com/lestrrat-go/jwx.

emilpriver:
Hey,

I totally agree with you after reading your comment. Better to use something that is built to have support for JWS rather then using something aimed for JWT to be used to create JWS. Closing the merge :D 

Thanks anyways for looking into the PR! :D

emilpriver:
Would it be a bad idea to build a JWS lib under the "golang-jwt" namespace? I did see that JWE exists: https://github.com/golang-jwt/jwe

oxisto:
> @oxisto Would it be a bad idea to build a JWS lib under the "golang-jwt" namespace? I did see that JWE exists: https://github.com/golang-jwt/jwe

I have been forwards and backwards on this myself. On the one hand, it would make sense to support the whole stack. On the other hand, we are very limited in terms of resources at the moment and are focused on bringing out v5 of this library. 

There are some good JWS libraries out there that support the whole stack, so it’s not a pressing issue at the moment.

emilpriver:
ok got ya! is this something that people outside go golang-jwt could build? If so then I would want to contribute to provide JWS for golang-jwt as well

PR #301, opened 2023-04-02
Support custom json and base64 encoders for Token and Parser

### LINKED ISSUE #168: Supporting alternative base64/json encoder/decoder

dcalsky:
This PR adds support for `Token` and `Parser` using custom JSON and base64 encoders to encode/decode related data for performance and experimentation purpose. 

By implementing the `Base64Encoder` interface with `EncodeToString` and `DecodeString` methods to create a custom base64 encoder, while implementing `JSONEncoder` with `Marshal` and `Unmarshal` methods to create a custom json encoder.

```go
import (
	"github.com/bytedance/sonic"
	asmbase64 "github.com/segmentio/asm/base64"
)

type sonicJsonEncoder struct{}

func (s *sonicJsonEncoder) Marshal(v any) ([]byte, error) {
	return sonic.Marshal(v)
}

func (s *sonicJsonEncoder) Unmarshal(data []byte, v any) error {
	return sonic.Unmarshal(data, v)
}

type asmBase64Encoder struct{}

func (s *asmBase64Encoder) EncodeToString(data []byte) string {
	return asmbase64.StdEncoding.EncodeToString(data)
}

func (s *asmBase64Encoder) DecodeString(data string) ([]byte, error) {
	return asmbase64.RawURLEncoding.DecodeString(data)
}
```

Then Using `WithJSONEncoder` and `WithBase64Encoder` to create a Parser, and using `WithTokenJSONEncoder` and `WithTokenBase64Encoder` to create a Token.

```go
// create the token and parser  with json encoder based on sonic and base64 encoder based on asm/base64
token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.RegisteredClaims{}, jwt.WithTokenJSONEncoder(&sonicJsonEncoder{}), jwt.WithTokenBase64Encoder(&asmBase64Encoder{}))

parser := jwt.NewParser(jwt.WithJSONEncoder(&sonicJsonEncoder{}), jwt.WithBase64Encoder(&asmBase64Encoder{}))
```

The disadvantages of this PR are (**Any suggestions are welcomed**):
- Introduce the third-party libs into the go module
- due to almost all of functions and structs are under the same package, the optional methods naming might be confused, like `WithJSONEncoder` and `WithTokenJSONEncoder`

 

Resolves #168


oxisto:
Very nice! I will do a review over the next days, but it’s basically the direction I envisioned. Just some general pointers: We want to avoid the inclusion of **any** third party libs, even if it’s only in the tests. I would suggest only testing the interface itself in the test without the actual sonic implementation.

dcalsky:
> Very nice! I will do a review over the next days, but it’s basically the direction I envisioned. Just some general pointers: We want to avoid the inclusion of **any** third party libs, even if it’s only in the tests. I would suggest only testing the interface itself in the test without the actual sonic implementation.

I've updated the PR according to your advice.

token_option.go
```diff
@@ -3,3 +3,15 @@ package jwt
 // TokenOption is a reserved type, which provides some forward compatibility,
 // if we ever want to introduce token creation-related options.
 type TokenOption func(*Token)
+
+func WithTokenJSONEncoder(enc JSONEncoder) TokenOption {
```
oxisto, line 7: Maybe name this WithJSONEncoder and the parse option WithJSONDecoder?

encoder.go
```diff
@@ -0,0 +1,13 @@
+package jwt
+
+// Base64Encoder is an interface that allows to implement custom Base64 encoding/decoding algorithms.
+type Base64Encoder interface {
```
oxisto, line 4: I wonder if it makes sense to split it into an Encoder and Decoder interface? And then name the options WithBase64Encoder and WithBase64Decoder for option and parser?

parser.go
```diff
@@ -12,17 +12,25 @@ type Parser struct {
 	// If populated, only these methods will be considered valid.
 	validMethods []string
 
-	// Use JSON Number format in JSON decoder.
+	// Use JSON Number format in JSON decoder. This field is disabled when using a custom json encoder.
```
oxisto, line 15:
```suggestion
	// Use JSON Number format in JSON decoder. This field is disabled when using a custom JSON encoder.
```

parser.go
```diff
@@ -135,7 +143,12 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		}
 		return token, parts, newError("could not base64 decode header", ErrTokenMalformed, err)
 	}
-	if err = json.Unmarshal(headerBytes, &token.Header); err != nil {
+	if p.jsonEncoder != nil {
+		err = p.jsonEncoder.Unmarshal(headerBytes, &token.Header)
```
oxisto, line 147: I think it would be cleaner to set `p.jsonEncoder` to `json.Unmarshal` (if nil) and then call `p.jsonEncoder.Unmarshal()` at the end like

```go
if p.jsonEncoder == nil {
  p.jsonEncoder = json.Unmarshal
}

err = p.jsonEncoder.Unmarshal(headerBytes, &token.Header)
```

parser.go
```diff
@@ -12,17 +12,25 @@ type Parser struct {
...
 	validator *validator
 
+	// This field is disabled when using a custom base64 encoder.
 	decodeStrict bool
 
+	// This field is disabled when using a custom base64 encoder.
 	decodePaddingAllowed bool
+
+	// Custom base64 encoder.
+	base64Encoder Base64Encoder
```
oxisto, line 30: maybe base64Decoder instead?

parser.go
```diff
@@ -12,17 +12,25 @@ type Parser struct {
...
+
+	// Custom json encoder.
+	jsonEncoder JSONEncoder
```
oxisto, line 33: maybe jsonDecoder (or Unmarshaller?) instead?

parser.go
```diff
@@ -146,21 +159,30 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
...
-	dec := json.NewDecoder(bytes.NewBuffer(claimBytes))
-	if p.useJSONNumber {
-		dec.UseNumber()
-	}
+
 	// JSON Decode.  Special case for map type to avoid weird pointer behavior
-	if c, ok := token.Claims.(MapClaims); ok {
-		err = dec.Decode(&c)
+	mapClaims, isMapClaims := token.Claims.(MapClaims)
+	if p.jsonEncoder != nil {
```
oxisto, line 165: Ok this is messy. It seems that maybe the better way would actually be to define an interface not for `Unmarshal` but for `Decode` and use `json.NewDecoder(....)` to fulfil this interface. 

Then we can basically re-use the old code that sets JSON number etc in case `jsonEncoder` (or decoder) is nil.

I am not quite sure how the other JSON libraries work. Do they also have a possible `Decoder` struct?

parser_option.go
```diff
@@ -125,3 +125,16 @@ func WithStrictDecoding() ParserOption {
 		p.decodeStrict = true
 	}
 }
+
+// WithJSONEncoder supports
+func WithJSONEncoder(enc JSONEncoder) ParserOption {
```
oxisto, line 130: Missing docstring

parser_option.go
```diff
@@ -125,3 +125,16 @@ func WithStrictDecoding() ParserOption {
...
+	return func(p *Parser) {
+		p.jsonEncoder = enc
+	}
+}
+
+func WithBase64Encoder(enc Base64Encoder) ParserOption {
```
oxisto, line 136: Missing docstring

token.go
```diff
@@ -14,12 +14,14 @@ type Keyfunc func(*Token) (interface{}, error)
...
-	Claims    Claims                 // Claims is the second segment of the token in decoded form
-	Signature []byte                 // Signature is the third segment of the token in decoded form.  Populated when you Parse a token
-	Valid     bool                   // Valid specifies if the token is valid.  Populated when you Parse/Verify a token
+	Raw           string                 // Raw contains the raw token.  Populated when you [Parse] a token
+	Method        SigningMethod          // Method is the signing method used or to be used
+	Header        map[string]interface{} // Header is the first segment of the token in decoded form
+	Claims        Claims                 // Claims is the second segment of the token in decoded form
+	Signature     []byte                 // Signature is the third segment of the token in decoded form.  Populated when you Parse a token
+	Valid         bool                   // Valid specifies if the token is valid.  Populated when you Parse/Verify a token
+	jsonEncoder   JSONEncoder            // jsonEncoder is the custom json encoder/decoder
```
oxisto, line 23: I know that the `Parser` has these fields directly embedded into the struct but in another PR we are trying to extract all the options into an unexported config options struct to not clutter the "main" struct. I would suggest doing this from the start and to move jsonEncoder and base64Encoder into a separate private config options struct.

token.go
```diff
@@ -64,12 +70,17 @@ func (t *Token) SignedString(key interface{}) (string, error) {
 // of the whole deal.  Unless you need this for something special, just go
 // straight for the SignedString.
 func (t *Token) SigningString() (string, error) {
-	h, err := json.Marshal(t.Header)
+	marshal := json.Marshal
```
oxisto, line 73: Move this into an else to avoid an assignment?

encoder.go
oxisto: It seems that at least https://github.com/bytedance/sonic#use-numberuse-int64 seems compatible to `encoding/json` so we could even support `UseNumber` with an interface like this:

```
type JSONDecoder interface {
  UseNumber()
  Decode(any) error
}
```

oxisto: Hmm it seems that using the decoder instead of `json.Unmarshal` is really slower (and takes slightly more allocations), so that's not a good idea. in this case maybe it would be worth it only to use NewDecoder if we really need it (e.g., if UseNumber is enabled) and only with the standard library.

parser.go
oxisto: See above. Using Decode(r) instead of Unmarshal should be avoided. Here is a quick suggestions for this code, it looks a bit hacky but does the trick a little bit better I think. I still need to run some numbers. I think this should be faster for most use-cases, where UseNumber is not needed:

```go
// JSON Decode.  Special case for map type to avoid weird pointer behavior
	mapClaims, isMapClaims := token.Claims.(MapClaims)

	var unmarshal func([]byte, any) error

	if p.jsonEncoder == nil {
		if p.useJSONNumber {
			unmarshal = func(data []byte, v any) error {
				decoder := json.NewDecoder(bytes.NewBuffer(claimBytes))
				decoder.UseNumber()
				return decoder.Decode(v)
			}
		} else {
			unmarshal = json.Unmarshal
		}
	} else {
		unmarshal = p.jsonEncoder.Unmarshal
	}

	if isMapClaims {
		err = unmarshal(claimBytes, &mapClaims)
	} else {
		err = unmarshal(claimBytes, &claims)
	}
```

alam-chime:
thanks @dcalsky 
+1 to this PR

oxisto:
Thanks for the initial work on this @dcalsky. I had another go at this, basically splitting up encoding and decoding and also converting most of the interfaces into function types. This should be pretty good integrated in the existing code base now. It even supports both parsing JSON with and without JSON number (as long as the underlying JSON library supports it).

I wonder whether I could also make our "special" base64 options (strict, allow padding) work with external libraries as well. For now, they are disabled when an external decoder is used.

If we are happy with this approach, I want to finalise docs and examples.

cc @mfridman @dcalsky

oxisto:
@mfridman Do we still want to pursue this? It did gather some interest from the community. If yes, I would do another cleanup pass and add docs and examples/tests.

mfridman:
I'll review this in the next few days.

mfridman:
It seems useful, albeit the number of requests for this was quite low.

I was mainly waiting to see what comes of https://github.com/golang/go/discussions/63397. If the native std lib could be improved then bringing your own encoding package becomes unnecessary?

I'm a bit on the fence on this one, but if we get it into a good enough shape I wouldn't block. My main hesitation is all the plumbing we have to do to support this in the current /v5 API.

yastrebtsov:
Hi,

I am waiting for this change because we are experiencing performance issues with large JWT tokens using the standard JSON parser.

> I was mainly waiting to see what comes of [golang/go#63397](https://github.com/golang/go/discussions/63397). If the native std lib could be improved then bringing your own encoding package becomes unnecessary?
According to their proposal, they are not going to improve performance significantly (like sonic json parser is doing).

Personally, I am waiting for this change or any other change which allows us to replace standard JSON parser.

Semerokozlyat:
Hello, would like to express an urgent demand in this change to boost up our JWT parsing logic on a highly loaded service.

oxisto:
> It seems useful, albeit the number of requests for this was quite low.
> 
> I was mainly waiting to see what comes of [golang/go#63397](https://github.com/golang/go/discussions/63397). If the native std lib could be improved then bringing your own encoding package becomes unnecessary?
> 
> I'm a bit on the fence on this one, but if we get it into a good enough shape I wouldn't block. My main hesitation is all the plumbing we have to do to support this in the current /v5 API.

@mfridman seems there is some more interest in this. Can you maybe give this review another go? I think it's not that much plumbing making it work. It is mainly two additional if-clauses in token and parse. The rest are just all optional options.

mfridman:
Thanks for the ping @oxisto, I'll take a pass this weekend. I get the notification but typically lack time during the week for more in-depth reviews.

dcalsky:
@oxisto Almost forget this PR... I will fine-tune this PR to ensure minimum changes today later or tomorrow.

dcalsky:
I have written some example tests using custom decoder and encoder. the new methods are very smooth and easy to use. It's time to merge them into the our main branch.

oxisto:
> I have written some example tests using custom decoder and encoder. the new methods are very smooth and easy to use. It's time to merge them into the our main branch.

Thanks! That also got rid of the pesky code coverage warning. I agree, I like the public interface part of this. I think @mfridman's concern is more about the inner workings of this inside parse/etc. I would abstain from further review comments now and give this into the hands of @mfridman since I heavily reworked this PR my self so I am not really neutral as a reviewer here :)

encoder.go
```diff
@@ -0,0 +1,35 @@
...
+import "io"
+
+// Base64Encoding represents an object that can encode and decode base64. A
+// common example is [encoding/base64.Encoding].
+type Base64Encoding interface {
+	EncodeToString(src []byte) string
+	DecodeString(s string) ([]byte, error)
+}
+
+type StrictFunc[T Base64Encoding] func() T
```
oxisto, line 12: I wonder if this type really needs to be exported at all

encoder.go
```diff
@@ -0,0 +1,35 @@
...
+
+type Stricter[T Base64Encoding] interface {
```
oxisto, line 14: This at least needs a comment if we want to keep It exported, it could probably also be private I guess

encoder.go
```diff
@@ -0,0 +1,35 @@
...
+	Strict() T
+}
+
+func DoStrict[S Base64Encoding, T Stricter[S]](x T) Base64Encoding {
```
oxisto, line 18: Double-check if this is actually still needed, which I think it is not?

encoder.go
```diff
@@ -0,0 +1,35 @@
...
+
+// JSONMarshalFunc is a function type that allows to implement custom JSON
+// encoding algorithms.
+type JSONMarshalFunc func(v any) ([]byte, error)
+
+// JSONUnmarshalFunc is a function type that allows to implement custom JSON
+// unmarshal algorithms.
+type JSONUnmarshalFunc func(data []byte, v any) error
+
+type JSONDecoder interface {
```
oxisto, line 30: Needs a comment because its a public interface

encoder.go
```diff
@@ -0,0 +1,35 @@
...
+	UseNumber()
+	Decode(v any) error
+}
+
+type JSONNewDecoderFunc[T JSONDecoder] func(r io.Reader) T
```
oxisto, line 35: Needs a comment

parser.go
```diff
@@ -12,16 +12,26 @@ type Parser struct {
...
 	// Skip claims validation during token parsing.
 	skipClaimsValidation bool
 
 	validator *Validator
 
-	decodeStrict bool
+	decoding
+}
+
+type decoding struct {
```
oxisto, line 23: its not private, but still we probably could use some comments here, especially for this outer struct

token.go
```diff
@@ -34,6 +34,13 @@ type Token struct {
 	Claims    Claims                 // Claims is the second segment of the token in decoded form
 	Signature []byte                 // Signature is the third segment of the token in decoded form.  Populated when you Parse a token
 	Valid     bool                   // Valid specifies if the token is valid.  Populated when you Parse/Verify a token
+
+	encoders
+}
+
+type encoders struct {
```
oxisto, line 41: its not private, but still we probably could use some comments here, especially for this outer struct

parser.go
```diff
@@ -160,25 +181,31 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
...
-	// the claims. However, this comes with a performance penalty so only use
-	// it if we must and, otherwise, simple use json.Unmarshal.
-	if !p.useJSONNumber {
-		// JSON Unmarshal. Special case for map type to avoid weird pointer behavior.
-		if c, ok := token.Claims.(MapClaims); ok {
-			err = json.Unmarshal(claimBytes, &c)
-		} else {
-			err = json.Unmarshal(claimBytes, &claims)
+	// If `useJSONNumber` is enabled, then we must use a dedicated JSONDecoder
+	// to decode the claims. However, this comes with a performance penalty so
```
oxisto, line 185: small nit: we only know definitly that the performance penalty occurs when using `json.NewDecoder`. Third-party libraries might not have that penalty, so we might at least state this like ... a performance penalty (when using the standard library decoder).

parser_option.go
```diff
@@ -121,8 +124,80 @@ func WithPaddingAllowed() ParserOption {
 // WithStrictDecoding will switch the codec used for decoding JWTs into strict
 // mode. In this mode, the decoder requires that trailing padding bits are zero,
 // as described in RFC 4648 section 3.5.
+//
+// Note: This is only supported when using [encoding/base64.Encoding], but not
```
oxisto, line 128: I think this note is not true anymore, since we are checking whether the library supports a `Strict()` function. We need to check this.

parser_option.go
```diff
@@ -121,8 +124,80 @@ func WithPaddingAllowed() ParserOption {
...
+// by any other decoder specified with [WithBase64Decoder].
 func WithStrictDecoding() ParserOption {
 	return func(p *Parser) {
 		p.decodeStrict = true
 	}
 }
+
+// WithJSONDecoder supports a custom JSON decoder to use in parsing the JWT.
+// There are two functions that can be supplied:
+//   - jsonUnmarshal is a [JSONUnmarshalFunc] that is used for the
```
oxisto, line 138: typo: double "the"

parser_option.go
```diff
@@ -121,8 +124,80 @@ func WithPaddingAllowed() ParserOption {
...
+		// This seems to be necessary, since we don't want to store the specific
+		// JSONDecoder type in our parser, but need it in the function
+		// interface.
+		p.jsonNewDecoder = func(r io.Reader) JSONDecoder {
+			return jsonNewDecoder(r)
+		}
+	}
+}
+
+// WithBase64Decoder supports a custom Base64 when decoding a base64 encoded
```
oxisto, line 168: grammer: "a custom Base64 [something missing here]"

token_option.go
```diff
@@ -3,3 +3,15 @@ package jwt
 // TokenOption is a reserved type, which provides some forward compatibility,
 // if we ever want to introduce token creation-related options.
 type TokenOption func(*Token)
+
+func WithJSONEncoder(f JSONMarshalFunc) TokenOption {
```
oxisto, line 7: missing comments here

token_option.go
```diff
@@ -3,3 +3,15 @@ package jwt
...
+	return func(token *Token) {
+		token.jsonMarshal = f
+	}
+}
+
+func WithBase64Encoder(enc Base64Encoding) TokenOption {
```
oxisto, line 13: missing comments here

parser.go
dcalsky: Any evidence?

oxisto: Yes. https://github.com/golang-jwt/jwt/issues/303

dcalsky:
Do we just miss the code annotation part now, if it is supplied, the PR would pass? @oxisto

parser.go
```diff
@@ -148,7 +158,18 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	if headerBytes, err = p.DecodeSegment(parts[0]); err != nil {
 		return token, parts, newError("could not base64 decode header", ErrTokenMalformed, err)
 	}
-	if err = json.Unmarshal(headerBytes, &token.Header); err != nil {
+
+	// Choose our JSON decoder. If no custom function is supplied, we use the standard library.
+	var unmarshal JSONUnmarshalFunc
+	if p.jsonUnmarshal != nil {
```
MikeYast, line 164: BTW, why are we determining whether a custom JSON marshaller is present in the Parse() method instead of during the Parser creation? The same applies to other places. Parser.jsonUnmarshal should be initialized during Parser construction, as we're not planning to replace JSON marshallers or base64 encoders on the fly.

parser.go
```diff
@@ -200,18 +227,37 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 // take into account whether the [Parser] is configured with additional options,
 // such as [WithStrictDecoding] or [WithPaddingAllowed].
 func (p *Parser) DecodeSegment(seg string) ([]byte, error) {
-	encoding := base64.RawURLEncoding
+	var encoding Base64Encoding
+	if p.rawUrlBase64Encoding != nil {
```
MikeYast, line 231: The same comment about stage where custom base64 encoder function set. We can set p.rawUrlBase64Encoding to  base64.RawURLEncoding or custom function on Parser creation

MikeYast: Yes, that sounds reasonable.

rjarmstrong:
Any updates on this PR - we are working on a service that requires jwt on the hot path and the std lib json unmarshal is impacting us a lot? Thanks

mfridman:
If someone wants to take a stab at a simpler implementation, I think we'd be open to merging an improvement.

I don't have the cycles atm, but looking through the PR it's quite complicated with all the generics stuff and could probably be simplified.

Gobd:
I normally write Go but have been doing some JS lately and used https://www.npmjs.com/package/fast-jwt. If you're parsing JWTs in the hot path maybe it's a good idea to try caching them like that package does. There is no point in doing the decoding and validation twice once it's already been done the result will always be the same. Just be careful not to cache longer than the validity of the JWT.

PR #302, opened 2023-04-02
A quick way to validate token string

dcalsky:
This PR reduces the scope of ToLower for strings and can provide a significant performance gain when validating Token Strings.

```go
import (
	"strings"
	"testing"
)

var (
	authA = "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyIxMjMiOiJka2FzbDtka2FzO2xka2FzbDtka2FzbDtka2FzbDtka2FsO3NrZGw7Y216eC4sbWMsLnp4bWMueix4bWMuLHp4bWMuengsbWMsLnp4bWMuLHp4bWMuLHp4bSxjLiIsImlhdCI6MTY4MDQ0Nzg2OX0.E9n6tm0P1-Z_fj7RmGpk6RpPMQzTJuo7knKNM0sBhvk"
)

func BenchmarkBearer(b *testing.B) {
	b.Run("previous", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			strings.HasPrefix(strings.ToLower(authA), "bearer ")
		}
	})
	b.Run("now", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			strings.HasPrefix(strings.ToLower(authA[:7]), "bearer ")
		}
	})
}

```

```
goos: darwin
goarch: arm64
pkg: awesomeProject
BenchmarkBearer
BenchmarkBearer/previous
BenchmarkBearer/previous-10         	 1675801	       719.9 ns/op
BenchmarkBearer/now
BenchmarkBearer/now-10              	40375887	        28.59 ns/op
```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -130,9 +130,6 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	// parse Header
 	var headerBytes []byte
 	if headerBytes, err = p.DecodeSegment(parts[0]); err != nil {
-		if strings.HasPrefix(strings.ToLower(tokenString), "bearer ") {
-			return token, parts, newError("tokenstring should not contain 'bearer '", ErrTokenMalformed)
-		}
 		return token, parts, newError("could not base64 decode header", ErrTokenMalformed, err)
 	}
 	if err = json.Unmarshal(headerBytes, &token.Header); err != nil {
--- a/request/extractor.go
+++ b/request/extractor.go
@@ -90,7 +90,7 @@ func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
 	tokenHeader := req.Header.Get("Authorization")
 	// The usual convention is for "Bearer" to be title-cased. However, there's no
 	// strict rule around this, and it's best to follow the robustness principle here.
-	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader), "bearer ") {
+	if len(tokenHeader) < 7 || !strings.HasPrefix(strings.ToLower(tokenHeader[:7]), "bearer ") {
 		return "", ErrNoTokenInRequest
 	}
 	return tokenHeader[7:], nil


oxisto:
Wow, to be honest I would not have thought that this is an issue. I wonder why the implementation of HasPrefix does not abort searching the string after n bytes. I will also look at this in the coming days if we have sufficient tests.

dcalsky:
> Wow, to be honest I would not have thought that this is an issue. I wonder why the implementation of HasPrefix does not abort searching the string after n bytes. I will also look at this in the coming days if we have sufficient tests.
> 
> Update: Nevermind, obviously the ToLower was the problem here

Hundreds of projects use this library for authentication, so I want to improve the performance of it as much as possible :D

oxisto (dismissed):

request/extractor.go
```diff
@@ -90,7 +90,7 @@ func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
 	tokenHeader := req.Header.Get("Authorization")
 	// The usual convention is for "Bearer" to be title-cased. However, there's no
 	// strict rule around this, and it's best to follow the robustness principle here.
-	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader), "bearer ") {
+	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader[:7]), "bearer ") {
```
mfridman, line 93: Don't we need a bounds check here, otherwise this will panic? How about:

```go
// getToken extracts the token from the request header. If the token is
// invalid, it returns an empty string.
func getToken(r *http.Request) string {
	tokenHeader := r.Header.Get("Authorization")
	if len(tokenHeader) > 7 && strings.ToLower(tokenHeader[0:6]) == "bearer" {
		return tokenHeader[7:]
	}
	return ""
}
```

dcalsky: Fixed it.

parser.go
```diff
@@ -130,7 +130,7 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	// parse Header
 	var headerBytes []byte
 	if headerBytes, err = p.DecodeSegment(parts[0]); err != nil {
-		if strings.HasPrefix(strings.ToLower(tokenString), "bearer ") {
+		if len(tokenString) < 7 || strings.HasPrefix(strings.ToLower(tokenString[:7]), "bearer ") {
```
oxisto, line 133: Nitpicking, but now this returns an error message that the token should not contain "bearer", if the token is too short, which might be confusing. Maybe like this instead?
```suggestion
		if len(tokenString) >= 7 && strings.HasPrefix(strings.ToLower(tokenString[:7]), "bearer ") {
```

I wonder if we even need this check at all. If the token really contains "bearer", the base64 decode will fail anyway, so maybe we could get rid of this alltogether?

dcalsky: > 

Remove them all and only keeps "could not base64 decode header" message is clearer and more efficiency, though it will lose some error information. I recommend this way.

oxisto (approved):

oxisto:
LGTM. Sorry for the long delay, I think we can ignore the 0.03 % coverage loss

cc @mfridman

mfridman (approved):

request/extractor.go
```diff
@@ -90,7 +90,7 @@ func (e BearerExtractor) ExtractToken(req *http.Request) (string, error) {
 	tokenHeader := req.Header.Get("Authorization")
 	// The usual convention is for "Bearer" to be title-cased. However, there's no
 	// strict rule around this, and it's best to follow the robustness principle here.
-	if tokenHeader == "" || !strings.HasPrefix(strings.ToLower(tokenHeader), "bearer ") {
+	if len(tokenHeader) < 7 || !strings.HasPrefix(strings.ToLower(tokenHeader[:7]), "bearer ") {
```
cbandy, line 93: This is late and a nitpick, but `HasPrefix` also slices the front. Is it faster to drop the call completely?

```suggestion
	if len(tokenHeader) < 7 || strings.ToLower(tokenHeader[:7]) != "bearer " {
```

oxisto: We adjusted it in a follow-up PR: https://github.com/golang-jwt/jwt/pull/329/files

PR #304, opened 2023-04-05
Added coverage reporting

oxisto:
This PR adds coverage reporting using coveralls.io. I still prefer codecov though, @mfridman do you have a preference?


mfridman:
I don't have a preference for one or the other, from my experience they add a lot of noise and giant comments to PR's.

I enabled access for Codecov, so this should now work... will try to push an empty commit.

mfridman (approved):
If you think this will help, lgtm.

PR #310, opened 2023-04-22
Replace HMAC with ED25519 in examples

G1gg1L3s:
While HMAC remains relatively popular in some contexts, it's generally considered as insecure, especially in untrustworthy environments, as it's easy for the verifier to forge the JWT token. [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html#jwt), for example, recommends using signature algorithms instead of HMAC.

Since ED25519 is state-of-the-art among signature schemes, this PR adds it to the examples. This is to ensure that the examples are "secure by default". As a result, users have less chances of using weak algorithms when they blindly copy the examples.

Though, I didn't touch the `hmac_example_test.go`, so if user really need HMAC tokens, they can explicitly go the corresponding examples.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,6 +1,8 @@
 package jwt_test
 
 import (
+	"crypto/ed25519"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"time"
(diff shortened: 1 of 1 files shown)


mfridman:
Should we have both examples? If we're missing ed25519 examples then for sure we should add them, but given how prevalent HMAC is it seems worth keeping?

oxisto:
The examples are a little bit of a mess since we have different ones in different files; I agree that we definitely want to have ones with priv/public key - maybe ECDSA as well. The question is also whether we really want to have them in the code repository anymore or on our documentation page, like here https://golang-jwt.github.io/jwt/usage/create/#with-additional-claims

G1gg1L3s:
> Should we have both examples? If we're missing ed25519 examples then for sure we should add them, but given how prevalent HMAC is it seems worth keeping?

Yes, this is exactly why I left the `hmac_example_test.go` untouched! Users can explicitly go for it if they really need. At the same time, the other examples provide algorithms more suitable for general use.

G1gg1L3s:
> The examples are a little bit of a mess since we have different ones in different files; I agree that we definitely want to have ones with priv/public key - maybe ECDSA as well. The question is also whether we really want to have them in the code repository anymore or on our documentation page, like here https://golang-jwt.github.io/jwt/usage/create/#with-additional-claims

I don't see problems having the examples here, considering also that they are runnable tests. But if you want, I would like to contribute to the examples on the https://golang-jwt.github.io!

oxisto:
> > The examples are a little bit of a mess since we have different ones in different files; I agree that we definitely want to have ones with priv/public key - maybe ECDSA as well. The question is also whether we really want to have them in the code repository anymore or on our documentation page, like here https://golang-jwt.github.io/jwt/usage/create/#with-additional-claims
> 
> I don't see problems having the examples here, considering also that they are runnable tests. But if you want, I would like to contribute to the examples on the https://golang-jwt.github.io!

Sure! You can find the docs repo here: https://github.com/golang-jwt/jwt-docs

example_test.go
```diff
@@ -14,24 +16,46 @@ import (
 // no way to retrieve other fields after parsing.
 // See the CustomClaimsType example for intended usage.
 func ExampleNewWithClaims_registeredClaims() {
-	mySigningKey := []byte("AllYourBase")
+	publicKey, privateKey, err := ed25519.GenerateKey(nil)
```
oxisto, line 19: I wonder whether we should load one of the keys from the `test` repository instead of generating one. This would probably be a bit more "realistic". Hopefully people will understand to load their own key instead of using the one in the test folder then.

G1gg1L3s: I saw that other tests load the key from the filesystem, so here I decided to show that they can also be generated if needed. At least it's secure and useful for quick prototypying.

PR #311, opened 2023-05-10
Fixed incorrect sign string (converted to bytes)

silverstr89:
there is a problem in case usage of string for signing. you have to convert it into slice bytes. simple fix will avoid this issues.

### DIFF
--- a/token.go
+++ b/token.go
@@ -52,6 +52,11 @@ func (t *Token) SignedString(key interface{}) (string, error) {
 		return "", err
 	}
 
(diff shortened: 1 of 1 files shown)


oxisto:
Hi!

Thank you for your contribution to this project. However, the usage of `[]byte` for the HMAC signing keys is intentional and we do not intent to change it to `string`. You can find the complete reasoning here in our documentation: https://golang-jwt.github.io/jwt/usage/signing_methods/#frequently-asked-questions

PR #312, opened 2023-06-01
Add error handling to examples

### LINKED ISSUE #309: Possible error in token parsing examples
I am using a modified version of the doc example [here](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#example-Parse-Hmac). My code is as follows:
```go
func VerifyJWT(secret string, tokenStr string) (string, int, error) {
	token, err := jwt.Parse(tokenStr, func(token *jwt.Token) (interface{}, error) {
		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
		}

		return []byte(secret), nil
	})

	if claims, ok := token.Claims.(jwt.MapClaims); ok && token.Valid {
                return claims["address"].(string), claims["user_id"].(int), nil
	}

	return "", 0, err
}
```

When given a bad tokenStr, the example breaks at the `if claims, ok := ...` clause and prints an error. However, I am getting a runtime panic:
```
runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:220 (0x404d4f5)
	panicmem: panic(memoryError)
/usr/local/go/src/runtime/signal_unix.go:818 (0x404d4c5)
	sigpanic: panicmem()
/Users/DIR/jwt.go:28 (0x45b8c85)
	VerifyJWT: if claims, ok := token.Claims.(jwt.MapClaims); ok && token.Valid {
```

However, the error is resolved when I test for `err != nil`:
```go
func VerifyJWT(secret string, tokenStr string) (string, int, error) {
	token, err := jwt.Parse(tokenStr, func(token *jwt.Token) (interface{}, error) {
		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
		}

		return []byte(secret), nil
	})

        // TEST FOR ERR
        if err != nil {
                return "", 0, err
        }

	if claims, ok := token.Claims.(jwt.MapClaims); ok && token.Valid {
                 return claims["address"].(string), claims["user_id"].(int), nil
	}

	return "", 0, err
}
```

I am testing with `secret="0x123456789"` and `tokenStr="hello"`. `err` is `token is malformed token contains an invalid number of segments`.

craigpastro:
The error after calling jwt.Parse is often omitted while other errors are checked. The lack of consistency can be a bit confusing I think, so this PR adds some of the checks in.

Resolves #309 I suppose.

### DIFF
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -146,15 +146,15 @@ func verifyToken() error {
 		return data, nil
 	})
 
+	// Print an error if we can't parse for some reason
+	if err != nil {
+		return fmt.Errorf("couldn't parse token: %w", err)
+	}
+
 	// Print some debug data
 	if *flagDebug && token != nil {
 		fmt.Fprintf(os.Stderr, "Header:\n%v\n", token.Header)
 		fmt.Fprintf(os.Stderr, "Claims:\n%v\n", token.Claims)
-	}
-
-	// Print an error if we can't parse for some reason
-	if err != nil {
-		return fmt.Errorf("couldn't parse token: %w", err)
 	}
 
 	// Is token invalid?
@@ -274,7 +274,7 @@ func showToken() error {
 	}
 
 	token, err := jwt.Parse(string(tokData), nil)
-	if token == nil {
+	if err != nil || token == nil {
 		return fmt.Errorf("malformed token: %w", err)
 	}
 
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -152,14 +152,9 @@ func verifyToken() error {
 	}
 
 	// Print some debug data
-	if *flagDebug && token != nil {
+	if *flagDebug {
 		fmt.Fprintf(os.Stderr, "Header:\n%v\n", token.Header)
 		fmt.Fprintf(os.Stderr, "Claims:\n%v\n", token.Claims)
-	}
-
-	// Is token invalid?
-	if !token.Valid {
-		return fmt.Errorf("token is invalid")
 	}
 
 	// Print the token details
@@ -274,7 +269,7 @@ func showToken() error {
 	}
 
 	token, err := jwt.Parse(string(tokData), nil)
-	if err != nil || token == nil {
+	if err != nil {
 		return fmt.Errorf("malformed token: %w", err)
 	}
 
--- a/example_test.go
+++ b/example_test.go
@@ -88,11 +88,11 @@ func ExampleParseWithClaims_customClaimsType() {
 		return []byte("AllYourBase"), nil
 	})
 	if err != nil {
-		fmt.Println(err)
+		log.Fatal(err)
 	} else if claims, ok := token.Claims.(*MyCustomClaims); ok {
 		fmt.Println(claims.Foo, claims.RegisteredClaims.Issuer)
 	} else {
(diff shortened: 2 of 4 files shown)


cmd/jwt/main.go
```diff
@@ -146,17 +146,17 @@ func verifyToken() error {
 		return data, nil
 	})
 
+	// Print an error if we can't parse for some reason
```
craigpastro, line 149: I think we should check the error first before trying other stuff so I moved this up.

example_test.go
```diff
@@ -86,6 +87,9 @@ func ExampleParseWithClaims_customClaimsType() {
 	token, err := jwt.ParseWithClaims(tokenString, &MyCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
 		return []byte("AllYourBase"), nil
 	})
+	if err != nil {
```
craigpastro, line 90: The `fatal` function is available here, and elsewhere, but without context it may be confusing, so I just expanded it.

oxisto (commented):
Thanks! It's a start, but I am still not really too happy about the whole situation where we have an `err` and `token.Valid`. They pretty much represent the same information, i.e., if there was no error, `token.Valid` will always be true. However, if there is an error, `token.Valid` is *most likely*. be true, but it could also be that the error happens before token validation, and thus `token` is `nil`.

I would even lean towards only checking for `err` and leaving out checks to `token.Valid`

craigpastro:
👋 @oxisto. Thanks for the review! That makes good sense. I removed all the `token.Valid` checks after checking the error.

example_test.go
```diff
@@ -86,8 +87,11 @@ func ExampleParseWithClaims_customClaimsType() {
...
+		log.Fatal(err)
+	}
 
-	if claims, ok := token.Claims.(*MyCustomClaims); ok && token.Valid {
+	if claims, ok := token.Claims.(*MyCustomClaims); ok {
```
oxisto, line 94: Wouldn't it be "cleaner" to have something like this:

```go
if claims, ok := token.Claims.(*MyCustomClaims); ok && err == nil {
  fmt.Printf("%v %v", claims.Foo, claims.RegisteredClaims.Issuer)
} else {
  fmt.Println(err)
}
```

or probably more elaborate like

```go
if err != nil {
   fmt.Println(err)
} else if claims, ok := token.Claims.(*MyCustomClaims); ok {
   fmt.Printf("%v %v", claims.Foo, claims.RegisteredClaims.Issuer)
} else {
   fmt.Printf("ParseWithClaims returned an unknown claims type, cannot proceed")
}
```

craigpastro: Absolutely. It didn't make much sense before printing out the error in the else clause as it would be nil. I updated to your latter example. How does it look?

oxisto: A regular `Println` would suffice, instead of `Printf`, but otherwise it looks good. Would you mind adjusting the other four occurences as well?

craigpastro: Oops. Yep, replaced `Printf` with `Println` where I could.

oxisto (approved):
I adjusted the use of else-if consistently, looks good then

mfridman (approved):

PR #323, opened 2023-06-29
Providing descriptive error message while dealing with invalid key(s)

parthokr:
Trying to [verify](https://github.com/golang-jwt/jwt/blob/b2b650971a1e882f0f3d4890b64db5f784b3e0c3/signing_method.go#L15) or [sign](https://github.com/golang-jwt/jwt/blob/b2b650971a1e882f0f3d4890b64db5f784b3e0c3/signing_method.go#L16C2-L16C6) with a key of type any, namely empty interface, on a token gives no hint about right key type. It would be great if we could return descriptive error message(s) (`ErrInvalidEd25519PublicKeyType`, `ErrInvalidEd25519PrivateKeyType`, `ErrInvalidRSAPublicKeyType`, `ErrInvalidRSAPrivateKeyType`, `ErrInvalidECDSAPublicKeyType`, `ErrInvalidECDSAPrivateKeyType`, and `ErrInvalidHMACKeyType`) with the right key type rather than raising `ErrInvalidKeyType` error for all types of signing and verifying functions.

Also removed unused `ErrTokenInvalidId` error from errors.go


errors.go
```diff
@@ -6,22 +6,27 @@ import (
 )
 
 var (
-	ErrInvalidKey                = errors.New("key is invalid")
-	ErrInvalidKeyType            = errors.New("key is of invalid type")
```
mfridman, line 10: This is a breaking change.

mfridman (changes requested):
Thank you for the PR, but we can't accept this as-is because it's a breaking change.

Indeed it'd be great if more information could be derived from the error, but there are probably a bunch of projects that are asserting for this sentinel error.

parthokr:
> Thank you for the PR, but we can't accept this as-is because it's a breaking change.
> 
> Indeed it'd be great if more information could be derived from the error, but there are probably a bunch of projects that are asserting for this sentinel error.

I see. Thanks for your response.

PR #324, opened 2023-07-05
Return ed25519 key types from their parse functions

twz123:
Instead of the generic crypto key types. This is done in the other parse functions as well. The code already does the type assertion.

### DIFF
--- a/ed25519_utils.go
+++ b/ed25519_utils.go
@@ -1,7 +1,6 @@
 package jwt
 
 import (
-	"crypto"
 	"crypto/ed25519"
 	"crypto/x509"
 	"encoding/pem"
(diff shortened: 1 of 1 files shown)


ed25519_utils.go
```diff
@@ -14,7 +13,7 @@ var (
 )
 
 // ParseEdPrivateKeyFromPEM parses a PEM-encoded Edwards curve private key
-func ParseEdPrivateKeyFromPEM(key []byte) (crypto.PrivateKey, error) {
+func ParseEdPrivateKeyFromPEM(key []byte) (ed25519.PrivateKey, error) {
```
mfridman, line 16: Given these are public functions, and the underlying types are different, I think this would be a breaking change.

We avoid breaking changes in major versions.

I do agree having parity would be nice, but not at the expense of introducing an incompatible change.

mfridman:
Closing for now, but we'll add this to https://github.com/golang-jwt/jwt/issues/335

PR #343, opened 2023-09-07
Add explicit ClaimsValidator implementation check for custom claims

epelc:
Prevent user from misnaming or fat fingering the Validate() method implementation.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -125,7 +125,7 @@ type MyCustomClaims struct {
 }
 
 // Ensure we implement [jwt.ClaimsValidator] at compile time so we know our custom Validate method is used.
-var _ jwt.ClaimsValidator = MyCustomClaims{}
+var _ jwt.ClaimsValidator = (*MyCustomClaims)(nil)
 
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -124,6 +124,9 @@ type MyCustomClaims struct {
 	jwt.RegisteredClaims
 }
 
+// Ensure we implement [jwt.ClaimsValidator] at compile time so we know our custom Validate method is used.
+var _ jwt.ClaimsValidator = MyCustomClaims{}
```
mfridman, line 128: nit, can we change this to `(*MyCustomClaims)(nil)` ?

mfridman (approved):
Thanks for adding this.

example_test.go
oxisto:
```suggestion
var _ jwt.ClaimsValidator = (*MyCustomClaims)(nil)
```

oxisto (approved):

oxisto (approved):

PR #344, opened 2023-09-09
Key rotation with VerificationKeySet

mfridman:
I wanted to see what this looked like by taking #328 and then adding a `PublicKeyset` with a well-defined set of keys. 

Most of this is recomendations from @oxisto and piggybacked on #328.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -93,9 +93,9 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
 	}
 
 	switch have := got.(type) {
-	case VerifyKeySet:
+	case VerificationKeySet:
 		if len(have.Keys) == 0 {
-			return token, newError("keyfunc returned empty keyset", ErrTokenUnverifiable)
+			return token, newError("keyfunc returned empty verification key set", ErrTokenUnverifiable)
 		}
 		// Iterate through keys and verify signature, skipping the rest when a match is found.
 		// Return the last error if no match is found.
--- a/parser.go
+++ b/parser.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"strings"
 )
 
@@ -81,34 +80,6 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
 		return token, newError("no keyfunc was provided", ErrTokenUnverifiable)
 	}
 
-	keys := make([]interface{}, 1)
-	// Convert the key or list of keys into a list of keys.
-	{
-		got, err := keyFunc(token)
-		if err != nil {
-			return token, newError("error while executing keyfunc", ErrTokenUnverifiable, err)
-		}
-
-		switch have := got.(type) {
-		case []interface{}:
-			keys = have
-		case []byte, []int8: // HMAC is an outlier, so treat it specially.
-			keys[0] = have
-		case interface{}:
-			typ := reflect.TypeOf(have)
-			switch typ.Kind() {
-			case reflect.Array, reflect.Slice:
-				val := reflect.ValueOf(have)
-				keys = make([]interface{}, val.Len())
-				for i := 0; i < val.Len(); i++ {
-					keys[i] = val.Index(i).Interface()
-				}
-			default:
-				keys[0] = have
-			}
-		}
-	}
-
 	// Decode signature
 	token.Signature, err = p.DecodeSegment(parts[2])
 	if err != nil {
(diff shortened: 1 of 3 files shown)


mfridman:
@oxisto can we tune coveralls threshold?

> coverage/coveralls — Coverage decreased (-0.02%) to 86.147%

Maybe if we can set it to 1% or something more sane?

parser.go
```diff
@@ -74,24 +74,40 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
...
+	got, err := keyFunc(token)
 	if err != nil {
-		return token, newError("could not base64 decode signature", ErrTokenMalformed, err)
+		return token, newError("error while executing keyfunc", ErrTokenUnverifiable, err)
 	}
 
-	// Perform signature validation
-	if err = token.Method.Verify(strings.Join(parts[0:2], "."), token.Signature, key); err != nil {
+	switch have := got.(type) {
+	case PublicKeyset:
```
mfridman, line 96: I suspect folks will try to use `[]jwt.PublicKey` here. What's the rationale for the "public key set" struct instead of asserting for a slice of PublicKey?

I guess it'd enable us to evolve the public key set in a backwards-compatible way if we wanted to attach additional fields.

oxisto:
> @oxisto can we tune coveralls threshold?
> 
> > coverage/coveralls — Coverage decreased (-0.02%) to 86.147%
> 
> Maybe if we can set it to 1% or something more sane?

Yeah -1 % or maybe -0.5% sounds reasonable

token.go
```diff
@@ -9,8 +10,22 @@ import (
...
 // Token.  This allows you to use properties in the Header of the token (such as
 // `kid`) to identify which key to use.
+//
+// The returned interface{} may be a single key or a PublicKeyset containing
+// multiple keys.
 type Keyfunc func(*Token) (interface{}, error)
 
+// PublicKey represents a generic public key interface.
+type PublicKey interface {
+	crypto.PublicKey | []uint8
```
oxisto, line 20: Should we also add `unsafeNoneMagicConstant` (for the `none` algorithm) to this? Then we would include all the allowed key types. Might make sense, if we ever change the `Keyfunc` function signature at some point.

token.go
```diff
@@ -9,8 +10,22 @@ import (
...
+}
+
+// PublicKeyset is a set of public keys that can be used to verify a token. It is used by the parser
+// to verify a token.
+type PublicKeyset struct {
```
oxisto, line 25: Keyset or KeySet? I always struggle with this as a non-native English speaker.

oxisto (commented):
I like the idea very much. Some smaller naming issues we should work out beforehand.

token.go
```diff
@@ -9,8 +10,22 @@ import (
 // the key for verification.  The function receives the parsed, but unverified
 // Token.  This allows you to use properties in the Header of the token (such as
 // `kid`) to identify which key to use.
+//
+// The returned interface{} may be a single key or a PublicKeyset containing
+// multiple keys.
 type Keyfunc func(*Token) (interface{}, error)
 
+// PublicKey represents a generic public key interface.
```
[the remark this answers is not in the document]
schmidtw: I think this would be the `DecodingKey` since it's the public key or the symmetric key.  The signing key is the private key generally.

schmidtw:
This is more clear than #328 .

token.go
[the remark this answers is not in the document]
oxisto: > I think this would be the "validating/decoding key" since it's the public key or the symmetric key. The signing key is the private key generally. Maybe `DecodingKey` or `ValidationKey`?

Ah yes of course.

parser.go
mfridman: To clarify, you're in favor of using a struct instead of a slice, correct?

token.go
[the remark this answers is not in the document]
mfridman: How about the term "verify" (either `VerifyKey` or `VerificationKey`), since this is where the keys are used? 

For example (using the shorter of the two):

```go
// VerifyKey represents a public or secret key for verifying a token's signature.
type VerifyKey interface {
	crypto.PublicKey | []uint8
}

// VerifyKeySet is a set of public or secret keys. It is used by the parser to verify a token.
type VerifyKeySet struct {
	Keys []VerifyKey
}
```

I typically prefer the term "verify" over "validate" when it comes to confirming authenticity. The term "validate" (in this codebase and more broadly) often covers a more broad set of checks.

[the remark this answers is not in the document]
oxisto: "verify" makes sense, since `SigningMethod` implements both `Sign` and `Verify` and this is actually the key used for `Verify`.

As for `VerifyKey` or `VerificationKey`: that's up to coin toss I guess ;)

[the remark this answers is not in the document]
mfridman: Pushed the renamed types in 1e49d85a02cca6c93007deceb16f240cfd5fd5a6

Given it's an interface, and the Go standard lib prefers nouns, the `VerificationKey` name is likely the most idiomatic. 

However, `VerifyKeySet` vs. `VerificationKeySet` is a coin toss. No strong opinion on this one, the former is shorter.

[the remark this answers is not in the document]
oxisto: I think having one "verify" and one "verification" is confusing, so I would probably go for "verification" on both then.

parser.go
```diff
@@ -74,24 +74,40 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
...
+		if len(have.Keys) == 0 {
+			return token, newError("keyfunc returned empty keyset", ErrTokenUnverifiable)
```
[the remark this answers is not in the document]
mfridman: Added an empty key set test to assert for a `ErrTokenUnverifiable` error.

token.go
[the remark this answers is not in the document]
mfridman: Updated, now we have parity:

```go
// VerificationKey represents a public or secret key for verifying a token's signature.
type VerificationKey interface {
	crypto.PublicKey | []uint8
}

// VerificationKeySet is a set of public or secret keys. It is used by the parser to verify a token.
type VerificationKeySet struct {
	Keys []VerificationKey
}
```

parser.go
oxisto: Yes

token.go
mfridman: I've been going back and forth on this one. Since `VerificationKey` is used in the key set and thus in verification, what does it mean to have a "rotating" key of none.

I wonder if this is something we can punt on and add when we do decide to expand the `VerificationKey` in more places? Afaics adding this later isn't a breaking change.

mfridman: Going to sleep on this one, and see if there's any feedback. But maybe you're right, adding it is the correct thing to do here so `VerificationKey` can represent all the keys (in a type-safe way) that we support.

oxisto: > Going to sleep on this one, and see if there's any feedback. But maybe you're right, adding it is the correct thing to do here so `VerificationKey` can represent all the keys (in a type-safe way) that we support.

Yes, that was my original thinking to have all of them in one place.

However, in re-reading the RFCs again, I am beginning to come to the conclusion that we might NOT even "need" to support the `none` algorithm. So this might also be the point in deciding to drop this in the future; or at least deprecate this in the future and NOT include this in this interface here.

mfridman: I think this PR accomplishes what we set out,

1. Add key rotation support
2. In a type-safe way
3. While adding a well-defined type for future improvements

My current thought is .. let's punt on adding (if ever) the `unsafeNoneMagicConstant` type. I'd like to understand what the practical cases are for it, and/or whether there is a true community need for this.

Starting with a narrow scope won't tie our hands in the future. But afaics we can expand the interface in the future if necessary.

oxisto (approved):

PR #349, opened 2023-09-27
Exported `NewValidator`

oxisto:
Previously, we had `newValidator` as a private function. This PR exports this function so that validation can be done independtly of parsing the claim.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -68,6 +68,16 @@ type Claims interface {
 }
 ```
 
+Users that previously directly called the `Valid` function on their claims,
+e.g., to perform validation independently of parsing/verifying a token, can now
+use the `jwt.NewValidator` function to create a `validator` independently of the
+`Parser`.
+
+```go
+var v = jwt.NewValidator(jwt.WithLeeway(5*time.Second))
+v.Validate(myClaims)
+```
+
 ### Supported Claim Types and Removal of `StandardClaims`
 
 The two standard claim types supported by this library, `MapClaims` and
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -17,7 +17,7 @@ and corresponding updates for existing programs.
 
 ## Parsing and Validation Options
 
-Under the hood, a new `validator` struct takes care of validating the claims. A
+Under the hood, a new `Validator` struct takes care of validating the claims. A
 long awaited feature has been the option to fine-tune the validation of tokens.
 This is now possible with several `ParserOption` functions that can be appended
 to most `Parse` functions, such as `ParseWithClaims`. The most important options
@@ -70,7 +70,7 @@ type Claims interface {
 
 Users that previously directly called the `Valid` function on their claims,
 e.g., to perform validation independently of parsing/verifying a token, can now
-use the `jwt.NewValidator` function to create a `validator` independently of the
+use the `jwt.NewValidator` function to create a `Validator` independently of the
 `Parser`.
 
 ```go
--- a/validator.go
+++ b/validator.go
@@ -28,13 +28,12 @@ type ClaimsValidator interface {
 	Validate() error
 }
 
-// validator is the core of the new Validation API. It is automatically used by
+// Validator is the core of the new Validation API. It is automatically used by
 // a [Parser] during parsing and can be modified with various parser options.
 //
-// Note: This struct is intentionally not exported (yet) as we want to
-// internally finalize its API. In the future, we might make it publicly
-// available.
-type validator struct {
+// The [NewValidator] function should be used to create an instance of this
+// struct.
+type Validator struct {
 	// leeway is an optional leeway that can be provided to account for clock skew.
 	leeway time.Duration
 
@@ -62,16 +61,28 @@ type validator struct {
 	expectedSub string
 }
 
-// newValidator can be used to create a stand-alone validator with the supplied
+// NewValidator can be used to create a stand-alone validator with the supplied
 // options. This validator can then be used to validate already parsed claims.
-func newValidator(opts ...ParserOption) *validator {
+//
+// Note: Under normal circumstances, explicitly creating a validator is not
+// needed and can potentially be dangerous; instead functions of the [Parser]
+// class should be used.
+//
+// The [Validator] is only checking the *validity* of the claims, such as its
+// expiration time, but it does NOT perform *signature verification* of the
+// token.
+func NewValidator(opts ...ParserOption) *Validator {
 	p := NewParser(opts...)
 	return p.validator
 }
 
 // Validate validates the given claims. It will also perform any custom
 // validation if claims implements the [ClaimsValidator] interface.
-func (v *validator) Validate(claims Claims) error {
+//
+// Note: It will NOT perform any *signature verification* on the token that
+// contains the claims and expects that the [Claim] was already successfully
+// verified.
+func (v *Validator) Validate(claims Claims) error {
 	var (
 		now  time.Time
 		errs []error = make([]error, 0, 6)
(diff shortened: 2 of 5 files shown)


oxisto:
I wonder if it would make sense to return an interface instead?

jcburley:
This worked perfectly (as far as building and passing unit tests in two of our repos), so I would like to see it Merged (and Released) if TPTB approve.

validator.go
```diff
@@ -62,9 +62,9 @@ type validator struct {
 	expectedSub string
 }
 
-// newValidator can be used to create a stand-alone validator with the supplied
+// NewValidator can be used to create a stand-alone validator with the supplied
 // options. This validator can then be used to validate already parsed claims.
-func newValidator(opts ...ParserOption) *validator {
+func NewValidator(opts ...ParserOption) *validator {
```
mfridman, line 67: I think the intent was to keep this unexported because it's a massive foot gun.

Should we add a warning comment directing folks to the "typical" usage?

mfridman:
> I wonder if it would make sense to return an interface instead?

Ye, it's uncommon to have an exported function return an unexported type.

An interface would be better, although then we'd be stuck with it in the public API.

jcburley:
What's the status of this? (We're "on hold" upgrading to jwt v5 pending resolution.)

mfridman:
I think it's "on hold' / "gathering more feedback", given this is a potential footgun.

We've been exporting and adding to the public API sparingly, and want to make sure we're at least having a discussion around it.

validator.go
oxisto: Just to be on the same page. What are your main concerns with it? I thought it is pretty much designed in a way that you cannot mess up default validation, but maybe I am too optimistic? ;)

mfridman: We were funnelling users towards the Parser, which did all the right things. Something about exposing a raw validator without verification seems "wrong". But I don't think that's a good argument for _not_ doing this.

TL;DR - ignore this point, and instead, let's focus on refactoring this to avoid retuning an unexported identifier `*validator`

oxisto: > We were funnelling users towards the Parser, which did all the right things. Something about exposing a raw validator without verification seems "wrong". But I don't think that's a good argument for _not_ doing this.

Ah, got it!

> 
> TL;DR - ignore this point, and instead, let's focus on refactoring this to avoid retuning an unexported identifier `*validator`

Yes, definitely. So we are either stuck exporting `*Validator` (without any public struct elements) or a new interface. Since the parser is also already a exported struct and not an interface, I would lean towards a struct as well for some consistency.

mfridman: Agreed, I was leaning towards a struct as well.

oxisto: I wonder if there is an alternative to directly exporting this somewhat "problematic" struct directly in the `jwt` package. Since its main use case are probably (unit) tests, I wonder if we could introduce a `jwttest` package, similar to the *test package in the stdlib, such as https://pkg.go.dev/net/http/httptest. We could then export the Validator (or maybe a TestValidator) in the `jwttest` package. To do so we probably would need to move the `validator` to an `internal` package.

Not sure if this is cleaner in the end but it would probably irritate people less than having the `Validator` and `Parser` struct directly in `jwt`. What do you think?

MIGRATION_GUIDE.md
```diff
@@ -68,6 +68,16 @@ type Claims interface {
 }
 ```
 
+Users that previously directly called the `Valid` function on their claims,
+e.g., to perform validation independently of parsing/verifying a token, can now
+use the `jwt.NewValidator` function to create a `validator` independently of the
```
jcburley, line 73:
```suggestion
use the `jwt.NewValidator` function to create a `Validator` independently of the
```

validator.go
mfridman: Indeed, this is a nice pattern (good idea). Let's keep this branch open, but open another to see how that would behave/feel.

I think there will always be a cohort of folks who'll want direct access to the Validator, but ideally, we gather a bit more feedback before committing to it.

jcburley: > I wonder if there is an alternative to directly exporting this somewhat "problematic" struct directly in the `jwt` package. Since its main use case are probably (unit) tests, I wonder if we could introduce a `jwttest` package, similar to the *test package in the stdlib, such as https://pkg.go.dev/net/http/httptest. We could then export the Validator (or maybe a TestValidator) in the `jwttest` package. To do so we probably would need to move the `Validator` to an `internal` package, so it can be accessed by us internally as well as by the `jwtttest` package.
> 
> Not sure if this is cleaner in the end but it would probably irritate people less than having the `Validator` and `Parser` struct directly in `jwt`. What do you think?

It's worth considering this at least, though I'm not sure `jwttest` is the best name for it.

Perhaps `jwtvalidator` or `jwtclaims`? (I was about to suggest `jwtonlyclaims`, but just thinking that name in my head made me kinda lol due to similarity to a popular website.)

vtolstov:
i'm realy need ability to validate generic claims and don't copy code from jwt repo.
What you want to provide if NewValidator does not exported?

validator.go
```diff
@@ -28,13 +28,12 @@ type ClaimsValidator interface {
...
+// Validator is the core of the new Validation API. It is automatically used by
 // a [Parser] during parsing and can be modified with various parser options.
 //
-// Note: This struct is intentionally not exported (yet) as we want to
-// internally finalize its API. In the future, we might make it publicly
-// available.
-type validator struct {
+// The [NewValidator] function should be used to create an instance of this
+// struct.
+type Validator struct {
```
mfridman, line 36: Another alternative (not sure if it's better/worse) is moving this to a public package, something like `package jwtvalidate`, which gets it out of the core `package jwt`.

We then document the intended use of _that_ package.

mfridman: What worries me is the example below. For those who understand what's happening, it's great. But there's going to be a cohort of folks that will conflate claim validation with token verification, and we'll be making this _extremely_ easy.

If someone assumes that checking claims alone is sufficient for token security, they might overlook the importance of verifying the token's integrity.

Having said that, I suppose since `ParseUnverified` is already the main offender, we're not accomplishing much by hiding the validator.

I hope these are valid concerns, and I'm not just being pedantic.

### Example

```go
rawToken := `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c`
var claims jwt.MapClaims
token, _, err := jwt.NewParser().ParseUnverified(rawToken, &claims)
if err != nil {
	log.Fatal(err)
}
if err := jwt.NewValidator().Validate(token.Claims); err != nil {
	log.Fatal(err)
}
fmt.Println(claims["name"])                               // John Doe
fmt.Println(time.Unix(int64(claims["iat"].(float64)), 0)) // 2018-01-17 20:30:22 -0500 EST
fmt.Println(claims["sub"])                                // 1234567890
```

validator.go
oxisto: I did a quick prototype on this, but it seems that it is impossible because of the way our types are currently structured. If we move the validator to an extra package, we still need to access types such as `Claims` (which resides in `jwt` from the `validator` package. In turn, we need to access `validator.Validator` then from the `Parser`, which leads to a cyclic dependency.

The only way to avoid that would probably to move all of the common types to a common (internal?) package and then use a lot of type aliases, but I am not sure if that is really a good solution for this problem.

In that case I would rather swallow the pill and let it reside in `jwt`.

validator.go
oxisto: See [above](https://github.com/golang-jwt/jwt/pull/349#discussion_r1386095589). Seems problematic because of cyclic dependencies.

validator.go
mfridman: > The only way to avoid that would probably to move all of the common types to a common (internal?) package and then use a lot of type aliases, but I am not sure if that is really a good solution for this problem.

That's what I was originally thinking, but, likely not worth it given the code complexity. Making the project harder to maintain isn't a good tradeoff.

mfridman (approved):

vtolstov:
waiting for release =)

oxisto:
> waiting for release =)

Unfortunately, we just released 5.1.0 so it might take a while as we usually wait until a few PRs come together. If you really need this feature you can use the pseudo version of the main branch / commit.

PR #350, opened 2023-09-27
Add deprecation notice to `Valid()`

oxisto:
This PR adds a deprecation notice to `Valid()` to inform users of our v4 branch, that direct usage of the function in a call is discouraged, since v5 changes the way the `Claims` interface looks like. This needs to be backported to the v4 branch and released as v4.5.1

See #348 for an extended discussion

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -9,6 +9,12 @@ import (
 // Claims must just have a Valid method that determines
 // if the token is invalid for any supported reason
 type Claims interface {
+	// Deprecated: Starting with v5, we will change the way the Claims interface
+	// work. This should only impact users that implemented their own claims
+	// struct (without embedding one of the supplied structs). Users of v4
+	// should already try to avoid calling this function directly. In most
+	// cases, it is sufficient to rely on the error messages of
+	// [Parser.ParseWithClaims] or to use the property [Token.Valid].
 	Valid() error
 }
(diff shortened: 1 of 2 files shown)


oxisto:
> Seems reasonable , although raises the question:
> 
> Is the /v4 version frozen and will only receive security and critical bug fixes?

I would say so. I am not sure if it makes sense to make a new patch release just with this deprecation notice. Maybe we wait until there is a real need for a release?

mfridman:
Tbh, I think that's what migration guides are for, or for an unreleased upcoming version that's in development. But retroactively adding a deprecation notice (to a "frozen" version) and cutting a release feels a bit off. 

Also, some users are perfectly content with the /v4 version and it doesn't make that part of the API "wrong".

oxisto:
> Tbh, I think that's what migration guides are for, or for an unreleased upcoming version that's in development. But retroactively adding a deprecation notice (to a "frozen" version) and cutting a release feels a bit off.
> 
> Also, some users are perfectly content with the /v4 version and it doesn't make that part of the API "wrong".
> 
> No blockers from me though, I trust your judgement and don't have a strong opinion on this one.

Makes sense! I have modified the migration guide in the PR dealing with the validation API, this seems to be a better approach.

PR #352, opened 2023-10-09
Add ID to Claims interface

### LINKED ISSUE #334: `GetID` in the `Claims` interface ?
Is there a special consideration for the absence of the `GetID()` function?

https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7

https://github.com/golang-jwt/jwt/blob/8aa5d6cef84fa859f10fed5db50bee48efc5f160/claims.go#L9-L16

st3fan:
This patch adds the `jti` claim to the `Claims` interface with a `GetID()` method. It also implements `GetID()` for both `MapClaims` and `RegisteredClaims`.

This makes it easier to use the registered claims (as documented in https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) when you don't know whether you have an instance of `RegisteredClaims` or `MapClaims` to work with.

(I have a a concrete use for this where it saves me a type switch to find out what type of claims I am working with)

I realize that this is a drive-by pull request - if this change is unwanted then please discard it.

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -5,7 +5,7 @@ package jwt
 // common basis for validation, it is required that an implementation is able to
 // supply at least the claim names provided in
 // https://datatracker.ietf.org/doc/html/rfc7519#section-4.1 namely `exp`,
-// `iat`, `nbf`, `iss`, `sub`, `aud` and `jti“.
+// `iat`, `nbf`, `iss`, `sub`, `aud` and `jti`.
 type Claims interface {
 	GetExpirationTime() (*NumericDate, error)
 	GetIssuedAt() (*NumericDate, error)
--- a/claims.go
+++ b/claims.go
@@ -5,7 +5,7 @@ package jwt
 // common basis for validation, it is required that an implementation is able to
 // supply at least the claim names provided in
 // https://datatracker.ietf.org/doc/html/rfc7519#section-4.1 namely `exp`,
-// `iat`, `nbf`, `iss`, `sub` and `aud`.
+// `iat`, `nbf`, `iss`, `sub`, `aud` and `jti“.
 type Claims interface {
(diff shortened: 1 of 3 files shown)


oxisto:
This looks like a good idea, unfortunately this is API breaking, so this is something we need to keep reserved for a future `v6`, so this might be "stuck" for a while.

claims.go
```diff
@@ -5,12 +5,13 @@ package jwt
...
-// `iat`, `nbf`, `iss`, `sub` and `aud`.
+// `iat`, `nbf`, `iss`, `sub`, `aud` and `jti`.
 type Claims interface {
 	GetExpirationTime() (*NumericDate, error)
 	GetIssuedAt() (*NumericDate, error)
 	GetNotBefore() (*NumericDate, error)
 	GetIssuer() (string, error)
 	GetSubject() (string, error)
 	GetAudience() (ClaimStrings, error)
+	GetID() (string, error)
```
oxisto, line 16: So in the spec this is called `JWT ID`, we tried to take the "name" of the claim of the spec and use that for the name of the function, so in that case it would be `GetJWTID()` although I am not really sure if this is a good name. Canonical Go would probably be `GetJwtId()`?

oxisto:
Let's do this, should be one of the first things that will land in `v6`.

claims.go
rodafr: All caps (i.e. JWTID) is idiomatic: https://go.dev/wiki/CodeReviewComments#initialisms

I think the even more idiomatic Go naming for these getters would actually be just Issuer(), Subject(), Audience(), ID(), etc.
Based on the Google Go styleguide, FWIW: https://google.github.io/styleguide/go/decisions#getters

PR #361, opened 2023-11-08
Improve ErrInvalidKeyType error message and Parse docs

Laurin-Notemann:
This small PR adds 
- a little bit more context to the ErrInvalidKeyType error message "key is of invalid go type (should be []byte())"
- better description of the Parse function

Generally the original error might seem obvious but for some reason it wasn't clear to me that it was specifically referring to the type in Go, I thought it was some kind of error with my jwt token and not my signing key. It seems like an easy fix of just improving the error message slightly so its clearer what is means.

Since I only looked at the example of how parse the jwt into a correct token again I was also really confused as to what "AllYourBase" even meant. Which is why I suggest to add a small comment, so its clear that this is referring to the signing key (which it only is if you read the previous examples and specifically look at the value of the signingKey).

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -7,7 +7,7 @@ import (
 
 var (
 	ErrInvalidKey                = errors.New("key is invalid")
-	ErrInvalidKeyType            = errors.New("key is of invalid type")
+	ErrInvalidKeyType            = errors.New("key is of invalid go type (should be []byte())")
 	ErrHashUnavailable           = errors.New("the requested hash function is unavailable")
 	ErrTokenMalformed            = errors.New("token is malformed")
 	ErrTokenUnverifiable         = errors.New("token is unverifiable")
--- a/example_test.go
+++ b/example_test.go
@@ -85,6 +85,7 @@ func ExampleParseWithClaims_customClaimsType() {
 	}
 
 	token, err := jwt.ParseWithClaims(tokenString, &MyCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
+    // []byte("AllYourBase") refers to the signing key that is used to sign the claim with
 		return []byte("AllYourBase"), nil
 	})
 	if err != nil {
@@ -109,6 +110,7 @@ func ExampleParseWithClaims_validationOptions() {
 	}
 
 	token, err := jwt.ParseWithClaims(tokenString, &MyCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
+    // []byte("AllYourBase") refers to the signing key that is used to sign the claim with
 		return []byte("AllYourBase"), nil
 	}, jwt.WithLeeway(5*time.Second))
 	if err != nil {
@@ -148,6 +150,7 @@ func ExampleParseWithClaims_customValidation() {
 	tokenString := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIiLCJpc3MiOiJ0ZXN0IiwiYXVkIjoic2luZ2xlIn0.QAWg1vGvnqRuCFTMcPkjZljXHh8U3L_qUjszOtQbeaA"
 
 	token, err := jwt.ParseWithClaims(tokenString, &MyCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
+    // []byte("AllYourBase") refers to the signing key that is used to sign the claim with
 		return []byte("AllYourBase"), nil
 	}, jwt.WithLeeway(5*time.Second))
 	if err != nil {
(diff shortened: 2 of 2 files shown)


errors.go
```diff
@@ -7,7 +7,7 @@ import (
 
 var (
 	ErrInvalidKey                = errors.New("key is invalid")
-	ErrInvalidKeyType            = errors.New("key is of invalid type")
+	ErrInvalidKeyType            = errors.New("key is of invalid go type (should be []byte())")
```
oxisto, line 10: This is problematic because it is used for all signing methods and the invalid type can be different ones, depending on the signing method. The only way to make this better is to add additional detail in the signing methods where the error is used.

Laurin-Notemann: Would you recommend to add a new Err into the Err list like so:
```
var (
	ErrInvalidKey                = errors.New("key is invalid")
	ErrInvalidKeyType            = errors.New("key is of invalid go type")
	...
 	ErrHmacKeyType               = errors.New("hmac expects []byte")
)
```
and then using it like so:
```
if !ok {
        return newError("", ErrInvalidKeyType, ErrHmacKeyType)
}
```

Or rather like this without creating a new Err in the list:
```
if !ok {
	return newError("hmac expects []byte", ErrInvalidKeyType)
}
```

Laurin-Notemann: Then I would add this for the rest of the Signing Methods as well, that should work then right?

oxisto: I think I would prefer the latter, I think its enough for people to check on the more general `ErrInvalidKeyType`, the specifics are fine as a string within `newError`.

rsa.go
```diff
@@ -73,7 +73,7 @@ func (m *SigningMethodRSA) Sign(signingString string, key interface{}) ([]byte,
 
 	// Validate type of key
 	if rsaKey, ok = key.(*rsa.PrivateKey); !ok {
-		return nil, ErrInvalidKey
+		return nil, newError("rsa sign expects *rsa.PrivateKey", ErrInvalidKeyType)
```
Laurin-Notemann, line 76: Not sure about this one, I wasn't quite sure what the reason is that this on is a ErrInvalidKey and not a ErrInvalidKeyType, but maybe there was is a reason.

rsa_pss.go
```diff
@@ -88,7 +88,7 @@ func (m *SigningMethodRSAPSS) Verify(signingString string, sig []byte, key inter
 	case *rsa.PublicKey:
 		rsaKey = k
 	default:
-		return ErrInvalidKey
+		return newError("rsapss verify expects *rsa.PublicKey", ErrInvalidKeyType)
```
Laurin-Notemann, line 91: same for this one..

rsa.go
oxisto: Yes, this should be ErrInvalidKeyType. I think the original idea was to differentiate between the wrong (go) type (ErrInvalidKeyType) and invalid parameters on the key itself (ErrInvalidKey), e.g., if the curve bits did not match, etc.

example_test.go
```diff
@@ -109,6 +110,7 @@ func ExampleParseWithClaims_validationOptions() {
 	}
 
 	token, err := jwt.ParseWithClaims(tokenString, &MyCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
+    // []byte("AllYourBase") refers to the signing key that is used to sign the claim with
```
oxisto, line 113: not sure if we really need a comment here to be honest.

errors.go
```diff
@@ -7,7 +7,7 @@ import (
 
 var (
 	ErrInvalidKey                = errors.New("key is invalid")
-	ErrInvalidKeyType            = errors.New("key is of invalid type")
+	ErrInvalidKeyType            = errors.New("key is of invalid go type")
```
oxisto, line 10: I would prefer to keep the error string the same of backwards compatibility. I think if we include the additional details, it is fairly obvious what this error is. If you want to stress the point from a documentation point of view, we can add a godoc here.

Laurin-Notemann: Yeah okay that makes sense I agree, since that error is only used in the context with the new information that should suffice

example_test.go
oxisto: We are somewhat "phasing" out those examples any way and want to build up our documentation website: https://golang-jwt.github.io/jwt/usage/parse/. Unfortunately, exactly the parsing aspect is not documented yet. I want to have that in the same style as https://golang-jwt.github.io/jwt/usage/create/, where each variable is explained in detail. If you want, you can of course contribute to the docs here: https://github.com/golang-jwt/jwt-docs/blob/main/docs/usage/parse.md

Laurin-Notemann: Oh that looks neat, I will look into that, I have reverted the changes for now then 👍 .

ecdsa.go
```diff
@@ -62,7 +62,7 @@ func (m *SigningMethodECDSA) Verify(signingString string, sig []byte, key interf
 	case *ecdsa.PublicKey:
 		ecdsaKey = k
 	default:
-		return ErrInvalidKeyType
+		return newError("ecdsa verify expects *ecsda.PublicKey", ErrInvalidKeyType)
```
oxisto, line 65: Minor thing. Should we uppercase `ECDSA` and the others so that it matches the signing method? Otherwise, it looks good.

oxisto (approved):
Looks good with a minor nit. Please also change the PR title and description to remove the parse docs.

oxisto (approved):

mfridman (approved):

PR #363, opened 2023-11-24
Update MIGRATION_GUIDE.md

jbarham:
Use correct path for v4 migration


oxisto:
Ah, good catch. That makes total sense. Thank you

oxisto (approved):

PR #366, opened 2023-12-14
Add Azp claim interface

mjudeikis:
This adds azp (authorized party) claim to the `Claims`. 

We ran into the issue where using OpenID this field comes up very often. It's not part of JWT spec, but it is part of OpenID (https://openid.net/specs/openid-connect-core-1_0.html#IDToken)  Especially in Azure AAD and Auth0. When doing nested OIDC providers. And we are not fully able to validate claims as depending on which authorization method you use they are mixing `aud` and `azp` fields 😿 

This is very similar to https://github.com/golang-jwt/jwt/pull/352

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -5,7 +5,7 @@ package jwt
 // common basis for validation, it is required that an implementation is able to
 // supply at least the claim names provided in
 // https://datatracker.ietf.org/doc/html/rfc7519#section-4.1 namely `exp`,
-// `iat`, `nbf`, `iss`, `sub` and `aud`.
+// `iat`, `nbf`, `iss`, `sub` and `aud`, as well as the optional `azp` claim.
 type Claims interface {
 	GetExpirationTime() (*NumericDate, error)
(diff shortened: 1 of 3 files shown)


mjudeikis:
cc: @oxisto

oxisto:
Unfortunately, we can not accept new claims to the `Claims` interface, because this is not backwards compatible. I would suggest using a custom claim that embeds one of the standard claims, as can be seen in this example: https://github.com/golang-jwt/jwt/blob/6bcdd9d5b6ecb03a80ac123d1a9dc363441cbffe/example_test.go#L37-L40

mfridman:
I think this class of issues is one of the most common. We should probably update the README.md and the docs site (maybe an FAQ) to explicitly call it out with a copy/paste example.

oxisto:
Closing this, as unfortunately we cannot add this into the standard claims interface. As mentioned above however it is fairly easy to just create a custom OpenID claim based on our JWT standard claims.

mjudeikis:
make sense :) thanks

Issue #392, opened 2024-06-21
Documentation around Parse() 


bt-nia:
The note in the docs (I guess sourced from the [comment here](https://github.com/golang-jwt/jwt/blob/v5.2.1/parser.go#L225))

> Parse parses, validates, verifies the signature and returns the parsed token. keyFunc will receive the parsed token and should return the cryptographic key for verifying the signature. The caller is strongly encouraged to set the WithValidMethods option to validate the 'alg' claim in the token matches the expected algorithm. For more details about the importance of validating the 'alg' claim, see https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

Sounds to me like the default of the library would be to just accept tokens with the none type alg. From my testing however, this does not seem to be the case:
```
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/golang-jwt/jwt/v5"
)

func main() {
	// Define a test user
	testUser := "testuser"

	// Define a test signing key
	secretKey := []byte("mysecretkey")

	// Create a new token using HS256 signing method
	tokenHS256 := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
		"user": testUser,
		"exp":  time.Now().Add(time.Hour * 1).Unix(),
	})

	// Create a new token using None signing method
	tokenNone := jwt.NewWithClaims(jwt.SigningMethodNone, jwt.MapClaims{
		"user": testUser,
		"exp":  time.Now().Add(time.Hour * 1).Unix(),
	})

	// Sign and get the complete encoded token as a string using the secret key
	tokenStringHS256, err := tokenHS256.SignedString(secretKey)
	if err != nil {
		log.Fatalf("Failed to sign token with HS256: %v", err)
	}

	// Sign and get the complete encoded token as a string using None algorithm
	tokenStringNone, err := tokenNone.SignedString(jwt.UnsafeAllowNoneSignatureType)
	if err != nil {
		log.Fatalf("Failed to sign token with None: %v", err)
	}

	// Print both JWTs
	fmt.Printf("HS256 Token: %s\n", tokenStringHS256)
	fmt.Printf("None Token: %s\n", tokenStringNone)

	// Validation tests
	fmt.Println("Test 1: Current Method")
	validateTokenCurrent(tokenStringHS256, secretKey)
	validateTokenCurrent(tokenStringNone, jwt.UnsafeAllowNoneSignatureType)

	fmt.Println("Test 2: HS256 Key by Default")
[...]

PR #393, opened 2024-06-21
adjust documentation of Parse()

bt-nia:
as discussed here:
https://github.com/golang-jwt/jwt/issues/392

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -215,13 +215,13 @@ func (p *Parser) DecodeSegment(seg string) ([]byte, error) {
 	return encoding.DecodeString(seg)
 }
 
-// Parse parses, validates, verifies the signature and returns the parsed token.
-// keyFunc will receive the parsed token and should return the cryptographic key
-// for verifying the signature. The caller is strongly encouraged to set the
-// WithValidMethods option to validate the 'alg' claim in the token matches the
-// expected algorithm. For more details about the importance of validating the
-// 'alg' claim, see
-// https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
+// Parse parses, validates, verifies the signature, and returns the parsed token. 
+// When a keyFunc is set and a tokenString is provided, the library will not accept 
+// the 'none' type algorithm, ensuring security by default. However, it is strongly 
+// recommended to explicitly specify the allowed algorithms using the WithValidMethods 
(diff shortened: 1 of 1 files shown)


parser.go
```diff
@@ -215,13 +215,13 @@ func (p *Parser) DecodeSegment(seg string) ([]byte, error) {
...
-// Parse parses, validates, verifies the signature and returns the parsed token.
-// keyFunc will receive the parsed token and should return the cryptographic key
-// for verifying the signature. The caller is strongly encouraged to set the
-// WithValidMethods option to validate the 'alg' claim in the token matches the
-// expected algorithm. For more details about the importance of validating the
-// 'alg' claim, see
-// https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
+// Parse parses, validates, verifies the signature, and returns the parsed token. 
+// When a keyFunc is set and a tokenString is provided, the library will not accept 
+// the 'none' type algorithm, ensuring security by default. However, it is strongly
```
oxisto, line 220: This is not entirely correct. By default no token is rejected based on an algorithm value alone. The specified algorithms is matched against a list of available algorithms `GetSigningMethod`:
https://github.com/golang-jwt/jwt/blob/62e504c2810b67f6b97313424411cfffb25e41b0/signing_method.go#L30

After this step, the key returned in `keyFunc` is forwarded to the signing method implementation, which checks whether the type of key is suitable for that particular method or not.

In order to avoid simple mistakes, the signing method of `none` only allows keys of a very special global variable `jwt.UnsafeAllowNoneSignatureType`.

bt-nia: I think I captured this, but just not that intricate detail. The main reason why I made the change here is that from the original comment it is not quite clear if the Parse function will be secure by default or not. It just mentions a rather unclear recommendation "caller is strongly encouraged to set the WithValidMethods option".

But hey, if you have a better idea for wording I'm all for it. I'm just looking to make the documentation clearer for users.

oxisto: Understood. I will try to come up with a very condensed version of what I wrote for the doc. The issue that I have with the wording "When a keyFunc is set and a tokenString is provided, the library will not accept the 'none' type algorithm, ensuring security by default" is that strictly speaking is not true. If you have a keyFunc that returns `jwt.UnsafeAllowNoneSignatureType` it will accept the 'none' type. The 'none' type does not have any special handling.

PR #396, opened 2024-07-04
Separate verification into its own function

quantumsheep:
Hello!

I have this use-case where I want to parse the JWT, fetch the secret elsewhere and then verify the JWT.

The current available functions forces me to parse the JWT another time. I want to validate the token, not parse it again.

Thanks,

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -58,6 +58,12 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
 		return token, err
 	}
 
+	return p.VerifyToken(token, parts, claims, keyFunc)
+}
+
+func (p *Parser) VerifyToken(token *Token, parts []string, claims Claims, keyFunc Keyfunc) (*Token, error) {
+	var err error
+
 	// Verify signing method is in the required set
 	if p.validMethods != nil {
 		var signingMethodValid = false


oxisto:
Maybe a trivial question, but can't you do this logic inside the keyfunc? You have access to the `*Token` which has been parsed (but not validated) up to this point. You can then fetch the secret and then return it. This is basically what libraries such as https://github.com/MicahParks/keyfunc do.

quantumsheep:
I sometimes only parse the JWT to speedup the process when it comes from a fully trusted source (from internal code). I could duplicate some code to make it work but separating the functions costs nothing and fits my use-case.

oxisto:
> I sometimes only parse the JWT to speedup the process when it comes from a fully trusted source (from internal code). I could duplicate some code to make it work but separating the functions costs nothing and fits my use-case.

We are extremely careful about introducing new public functions because we need to maintain them in a way that we cannot break their function signature for quite a long time (since we tend to stick with major versions for quite a while). So yes, separating these functions actually does costs something: the time of a maintainer ;)

We intentionally did not expose any of these functions to not confuse people who might not be as experienced as you and might be confused, whether a simple `Parse` is enough or if `VerifyToken` is also needed; probably further complicated through the fact that we also have now a "validator".

As a bare minimum this function needs a godoc string and we probably would need to have an additional though about its function signature, because as I said before we need to stick with it for quite a while.

oxisto:
any though on this @mfridman ?

mfridman:
Yeah I'm okay with this. Needs a godoc comment though.

parser.go
```diff
@@ -58,6 +58,12 @@ func (p *Parser) ParseWithClaims(tokenString string, claims Claims, keyFunc Keyf
 		return token, err
 	}
 
+	return p.VerifyToken(token, parts, claims, keyFunc)
+}
+
+func (p *Parser) VerifyToken(token *Token, parts []string, claims Claims, keyFunc Keyfunc) (*Token, error) {
```
oxisto, line 64: @quantumsheep Can you supply a good for this function? Then we can accept the merge.

oxisto:
Having another look at this I would actually advise against it. We already have too many Parse/ParseUnverified functions plus the validator. Exposing another function that only does the token verification. I'd rather not introduce anything new at this point and try to focus whether we could improve the API interface in v6 and possibly introduce a `Verify` function on the token itself (which may also do or not do the validation).

PR #398, opened 2024-07-10
add parser option to skip not before verification

sschulz-t:
This pr adds a WithoutNotBeforeVerification() ParserOption.

This functionality is needed when you want to verify a token that you received but that is not yet valid. 
One exemplary use case could be a cinema ticket that is valid tomorrow. You receive it today, verify it and want to add it to your wallet. 

I also added test cases. Let me know if something else needs to be changed.

### DIFF
--- a/parser_option.go
+++ b/parser_option.go
@@ -126,3 +126,9 @@ func WithStrictDecoding() ParserOption {
 		p.decodeStrict = true
 	}
 }
+
+func WithoutNotBeforeVerification() ParserOption {
+	return func(p *Parser) {
+		p.validator.skipNbfVerification = true
+	}
+}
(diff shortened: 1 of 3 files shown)


oxisto:
Thanks for your contribution. I am a little bit torn with this "feature". The standard explicitly mentions that tokens must only be processed when the date is after "not before". While we have a feature to completely disabled validation (which I do not like anyway), I am not sure if disabling bits and pieces of the standard one-by-one is a good idea.

Personally, in this use case I would separate the validity date from the `nbf`. There are some discussion about the general use-fulness of `nbf` in the first place. I would probably just use an application specific claim and then you are free to do your own logic on this. To give you an example: even though the ticket is not "valid" yet because the movie has not started, you might need to verify the validity of the token itself, e.g., to check that someone has probably paid for a ticket, etc.

sschulz-t:
I can see you concerns and i agree. This could lead to bad things if not used with caution. I will close this PR :)

PR #406, opened 2024-08-28
Fix jwt -show

AlexanderYastrebov:
Some time ago (was lazy to dig history) jwt.Parse was changed to require non-nil keyFunc.

This broke `jwt -show` command that prints header and claims and does not validate the token:
```
$ echo $TOKEN | jwt -show -
Error: malformed token: token is unverifiable: no keyfunc was provided
```

This change fixes `jwt -show`.

### DIFF
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -273,7 +273,7 @@ func showToken() error {
 		fmt.Fprintf(os.Stderr, "Token len: %v bytes\n", len(tokData))
 	}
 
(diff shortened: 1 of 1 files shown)


oxisto:
Sounds reasonably to me. Can we somehow add a little more documentation to the to itself that the token is only displayed but not verified?

oxisto (approved):

AlexanderYastrebov:
I've updated usage texts.

mfridman (approved):

PR #410, opened 2024-10-13
spelling: compatibility

jsoref:


equalsgibson:
@mfridman / @oxisto - can this be reviewed and merged? It's a one line change for a typo that was made

mfridman:
I think I looked at this and it was some automated thing. Didn't prioritize it. Would rather see something in CI that'll catch these errors which solves it long-term than a one-off here and there (which we get a lot of).

equalsgibson:
That makes sense - I hadn't considered it from that point of view. It looks like there are a few options that could solve this moving forward and catch typos.

I created a PR linked to this issue that automatically runs a Github Action to spellcheck. I've configured the CLI version and tested it locally, and get the following output after additional configuration:  
```
error: `compatibilty` should be `compatibility`
  --> ./VERSION_HISTORY.md:100:487
    |
100 | The second update, while unfortunately requiring a small change in integration, is required to open up this library to other signing methods.  Not all keys used for all signing methods have a single standard on-disk representation.  Requiring `[]byte` as the type for all keys proved too limiting.  Additionally, this implementation allows for pre-parsed tokens to be reused, which might matter in an application that parses a high volume of tokens with a small set of keys.  Backwards compatibilty has been maintained for passing `[]byte` to the RSA signing methods, but they will also accept `*rsa.PublicKey` and `*rsa.PrivateKey`.
    |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       ^^^^^^^^^^^^
    |
error: `unecessary` should be `unnecessary`
  --> ./parser.go:219:13
    |
219 |  // causing unecessary overhead parsing tokens.
    |             ^^^^^^^^^^
    |
```

More info on the action is here; https://github.com/crate-ci/typos/tree/master

Let me know if this is acceptable as a solution, or if you think more can be done. It seems you can enable the action to "write" and fix the typos, but IMO it would be better to flag them and let the author of the PR either add it to the config file as an exclusion (.github/workflows/spell-check.yml) or fix the spelling manually.

Issue #411, opened 2024-10-17
Wrong `verifyIssuedAt ` parameter


sknr:
https://github.com/golang-jwt/jwt/blob/6bcdd9d5b6ecb03a80ac123d1a9dc363441cbffe/validator.go#L115-L120

When using the `WithIssuedAt` ParserOption, the internal verifyIssuedAt method must have the **required** parameter set to `true` instead of `false` otherwise the option only recognises if the token is issued in the future but not if the `iat` claim is missing.

oxisto:
> https://github.com/golang-jwt/jwt/blob/6bcdd9d5b6ecb03a80ac123d1a9dc363441cbffe/validator.go#L115-L120
> 
> When using the `WithIssuedAt` ParserOption, the internal verifyIssuedAt method must have the **required** parameter set to `true` instead of `false` otherwise the option only recognises if the token is issued in the future but not if the `iat` claim is missing.

This is probably bit of a philosophical question. The usage of this claim is completely optional according to the standard and so is its verification. So the question is what does the user expect here and this probably comes down to two-use cases:

1. Validate the issued-at date, if it exists
2. Require and the issued-at date and validate it

Currently, we chose to implement option 1, however I can see some benefit in option 2. What I would suggest is to have to flags: `WithIssuedAt` only does the validation and `WithIssuedAtRequired` (we have a similar flag for expired) controls whether it is required or not. We probably would also need to adjust the documentation to make it clearer.

Would you be open to implement a PR for this?

PR #412, opened 2024-10-17
Adjust behaviour of WithIssuedAt ParserOption (#411)

sknr:
Hi @oxisto, 

thanks for your quick reply to my Issue (#411). 

I created a fix with a handling similar to verifyExpiresAt which is what I also expected from verifyIssuedAt.

Hope that makes sense to you. 

Let me know, if you have any further questions.

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -112,11 +112,10 @@ func (v *Validator) Validate(claims Claims) error {
 		errs = append(errs, err)
 	}
 
-	// Check issued-at if the option is enabled
-	if v.verifyIat {
-		if err = v.verifyIssuedAt(claims, now, false); err != nil {
-			errs = append(errs, err)
-		}
+	// Check issued-at if the option is enabled, but usage of the claim
+	// itself is OPTIONAL.
+	if err = v.verifyIssuedAt(claims, now, v.verifyIat); err != nil {
+		errs = append(errs, err)
 	}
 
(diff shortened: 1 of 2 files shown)


oxisto:
I am a bit worried about changing the default behaviour here. As I see it we have two options to enable this:
1. Change the default to the claim to be required when we use `WithIssuedAt`, this would be in line with the other `WithIssuer`, `WithAudience`, etc., which all switch to the claim to be REQUIRED if you specify them
2. Keep the function as is and introduce a new flag `WithIssuedAtRequired`, that only controls the required aspect. We already have a `WithExpiredAtRequired`, so this would be in line with this option, but sort of deviates from the others

There is probably not a good 100 % solution without too much breakage. This calls for some input form our specialist for maintaining API stability @mfridman ;)

sknr:
@oxisto, @mfridman,

Thank you for sharing your thoughts. Since I can’t implement solution one, I’d prefer not to add another method to the public API and risk polluting it further. Therefore, I’ll handle the case of a nil “iat” claim internally instead of introducing a new flag. I hope that works for everyone. Let me know if you have any concerns!

mfridman:
Thanks for opening an issue. We'll think through these little details if we ever get around to another major version.

We take backwards compatibility fairly seriously, and appreciate your understanding.

Issue #413, opened 2024-10-18
`ParseUnverified` doesn't set the token `Signature` field


slickwilli:
I'm not sure if this was an intentional design choice or just an oversight but `ParseUnverified` does not currently set the `token.Signature` field. 

Logically, I was expecting this field to still be set even though it is not verified. Extracting it manually is simple enough but I believe `ParseUnverified` should be functionally equivalent (minus verification).

I have a PR ready to go if this is indeed an oversight.

oxisto:
Implemented in https://github.com/golang-jwt/jwt/pull/414

PR #414, opened 2024-10-18
Set token.Signature in `ParseUnverified`

slickwilli:
Sets the token's `Signature` field in `ParseUnverified` as discussed in #413

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -196,6 +196,12 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		return token, parts, newError("signing method (alg) is unspecified", ErrTokenUnverifiable)
 	}
 
+	// Set token signature
+	token.Signature, err = p.DecodeSegment(parts[2])
+	if err != nil {
+		return token, newError("could not base64 decode signature", ErrTokenMalformed, err)
+	}
+
 	return token, parts, nil
 }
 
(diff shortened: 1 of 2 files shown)


parser.go
```diff
@@ -193,6 +193,12 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		return token, parts, newError("signing method (alg) is unspecified", ErrTokenUnverifiable)
 	}
 
+	// Set token signature
+	token.Signature, err = p.DecodeSegment(parts[2])
```
oxisto: Leaving this here since I can't comment on the lines above we should probably remove the corresponding part from the regular parse because that calls `ParseUnverified` and we decode it twice now.

oxisto:
> No issues from my end.

Great. I will adjust this PR to avoid the double-setting of the signature though

oxisto (approved):
@mfridman please have another look at my optimised version

mfridman (approved):

samfoxcode:
@mfridman @oxisto - previously `ParseUnverified` did not validate the signature, but with this PR the signature is being validated by way of base64 and if the signature is not valid base64 it will error. Since this differs from the documented behavior of the method (`ParseUnverified parses the token but **does not** validate the signature`) should this be published as a breaking change and not a patch? Currently this was pushed in `5.3.1` as a patch but this is an unexpected behavior change that now causes an error to be returned.

I've opened an issue here as well: https://github.com/golang-jwt/jwt/issues/499

mfridman:
Thanks @samfoxcode we'll take a look!

PR #420, opened 2024-11-23
Add token versioning

iraj720:
i have added versioning to the library. i think it makes it easier since we are facing this challenge.

also it doesn't effect other parts of the projects.

Thanks.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -98,6 +98,75 @@ func ExampleParseWithClaims_customClaimsType() {
 	// Output: bar test
 }
 
+type claimsV1 struct {
+	jwt.RegisteredClaims
+	ID  string
+	Exp int64
+}
+
+type claimsV2 struct {
+	jwt.RegisteredClaims
+	ID     string
+	UserID string
+}
+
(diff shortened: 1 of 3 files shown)


melissamforbs (approved):
v

melissamforbs (approved):
b

iraj720:
@melissamforbs could you please mention another maintainer so we can merge it?

Thanks

oxisto:
This seems to be a very specific use-case and personally I do not see a good reason to support something like this. You can always have a claims structure that looks like this (roughly)

```
type versionedClaims struct {
   jwt.RegisteredClaims
   V1 struct { /*... */ }
   V2 struct { /*... */ }
}
```

and then override `UnmarshalJSON` and `MarshalJSON` to fill the V1 or V2 properties based on an arbitrary property. The RFC itself does not specify any "version" variable or something similar.

mfridman:
Thanks for the PR, but this is a fairly specific use case, and I don't think it belongs in the core library.

Issue #342, opened 2023-09-04
Verifying multiple audiences


oberlage:
Hi there,

I've just updated to `v5` and found that the new `RegisteredClaims` (in `registered_claims.go`) struct allows for a `[]string` type via the `ClaimStrings` type. This opens the way for verifying multiple audiences from the token.

I'm very happy with this, as my authentication provider does provide multiple audiences and `v4` gave no option to verify these.

However, the new `ParserOption` named `WithAudience(aud string)` and accompanying `validator` still only allows for a single `string` audience to be verified.

With [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) specifically mentioning multiple audiences, it does feel like something nice to support.
> "(..) In the general case, the "aud" value is an array of case-sensitive strings (..)" 

My questions are as follows:
1. Are there specific reasons to not implement checking of multiple audiences?
2. Are there any plans to do so?
3. If not, how do you feel about having an additional  `ParserOption` called `WithAudiences(auds []string)` for the specific case of multiple audiences. This would not break the existing `WithAudience(..)` and add functionality.

Thanks in advance!

oxisto:
Just to double-check: Are you sure you want to check against multiple *expected* audiences? 

What `WithAudience` does, it checks *all* supplied audience, whether one of them includes the expected one (or is equal to it, in the case of just one audience). Usually your application should define one expected audience (e.g., its own hostname) and then check for it. I am  not so sure about what the semantics about multiple expected audiences would entail.

If there is a valid use case for it, I suppose we (or rather you ;) as part of a PR) could add the `WithAudiences` parser option. Although there is probably a discussion then whether all expected audiences must match or any of them.

We tried to do the most basic functionality first in `v5` and then see where we could add additional features that make sense.

chamilto:
@oxisto Would you still be open to accepting a PR for this feature? 

> Although there is probably a discussion then whether all expected audiences must match or any of them.

Perhaps the option could be configurable in this sense, e.g.

`func WithAudiences(auds []string, matchAll bool) ParserOption {}`

iRbouh:
Having multiple audiences is beneficial when using Google OAuth, especially since we have different apps across web and mobile, each with its client ID (audience). The server endpoint needs to validate against these multiple audiences.

Source: https://developers.google.com/identity/gsi/web/guides/verify-google-id-token

[...]

PR #426, opened 2024-12-17
allow multiple audiences

gruberlu:
Hey there!

This change allows for multiple allowed audiences when verifying a JWT (see #342 ).

I did not implement the suggested function signature `func WithAudiences(auds []string, matchAll bool) ParserOption {}` as this would either cause breaking changes or lead to confusion and unnecessary complexity as a separate option.

### DIFF
--- a/parser_option.go
+++ b/parser_option.go
@@ -66,15 +66,15 @@ func WithExpirationRequired() ParserOption {
 	}
 }
 
-// WithAudience configures the validator to require the specified audience in
-// the `aud` claim. Validation will fail if the audience is not listed in the
-// token or the `aud` claim is missing.
+// WithAudience configures the validator to require ONE of the specified
+// audiences to be present in the `aud` claim. Validation will fail if none of
+// the audiences is listed in the token or the `aud` claim is missing.
 //
 // NOTE: While the `aud` claim is OPTIONAL in a JWT, the handling of it is
 // application-specific. Since this validation API is helping developers in
 // writing secure application, we decided to REQUIRE the existence of the claim,
 // if an audience is expected.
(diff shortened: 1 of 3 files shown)


oxisto:
> Hey there!
> 
> This change allows for multiple allowed audiences when verifying a JWT (see #342 ).
> 
> I did not implement the suggested function signature `func WithAudiences(auds []string, matchAll bool) ParserOption {}` as this would either cause breaking changes or lead to confusion and unnecessary complexity as a separate option.

Can you elaborate how this would be a breaking change?

gruberlu:
You would either have two functions then (`WithAudience` and `WithAudiences`) which does not cause a breaking change but is a little confusing in my opinion or just have the new function `WithAudiences` which does break existing code.

parser_option.go
```diff
@@ -66,15 +66,15 @@ func WithExpirationRequired() ParserOption {
 	}
 }
 
-// WithAudience configures the validator to require the specified audience in
-// the `aud` claim. Validation will fail if the audience is not listed in the
-// token or the `aud` claim is missing.
+// WithAudience configures the validator to require ONE of the specified
```
shubham1172, line 69: A `matchAll` option would be useful here for users who want to ensure ALL audiences are present. 
https://github.com/golang-jwt/jwt/issues/342#issuecomment-2295309922

mfridman: I believe https://github.com/golang-jwt/jwt/pull/427 does just that, likely the implementation we'll go with.

shubham1172: I just saw that, thanks!

gruberlu:
closed in favour of #433

PR #427, opened 2024-12-24
HdonCr/Allow multiple audiences

hdonCr:
Allow for multiple expected audiences as requested in [issue 342](https://github.com/golang-jwt/jwt/issues/342).


oxisto:
Hi! Looks like we have two PRs now for the same issue :) I will take some time after the holidays with @mfridman to see which direction we will follow.

validator.go
```diff
@@ -55,6 +55,13 @@ type Validator struct {
 	// string will disable aud checking.
 	expectedAud string
 
+	//expectedAuds contains the audiences this token expects. Supplying an empty
+	// []string will disable auds checking.
+	expectedAuds []string
```
oxisto, line 60: I wonder whether we could merge this with the `expectedAud` above

validator.go
```diff
@@ -55,6 +55,13 @@ type Validator struct {
...
+
+	// expectedAudsMatchAll specifies whether all expected audiences must match all auds from claim
+	expectedAudsMatchAll bool
```
oxisto, line 63: I would prefer a name like `matchAllAud`

validator.go
```diff
@@ -126,6 +133,13 @@ func (v *Validator) Validate(claims Claims) error {
 		}
 	}
 
+	// If we have expected audiences, we also require the audiences claim
+	if len(v.expectedAuds) != 0 {
```
oxisto, line 137: should be `> 0` just to be safe

validator.go
```diff
@@ -255,6 +269,102 @@ func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) err
...
+//
+// If matchAllAuds is true and aud length does not match cmps length, an ErrTokenInvalidAudience error will be returned.
+// Note that this does not account for any duplicate aud or cmps
+//
+// If aud is not set or an empty list, it will succeed if the claim is not required,
+// otherwise ErrTokenRequiredClaimMissing will be returned.
+//
+// Additionally, if any error occurs while retrieving the claim, e.g., when its
+// the wrong type, an ErrTokenUnverifiable error will be returned.
+func (v *Validator) verifyAudiences(claims Claims, cmps []string, required bool, matchAllAuds bool) error {
```
oxisto, line 284: This function seems to be very complex, I wonder if we could streamline this implementation a little bit. Furthermore, it should be possible to include the previous implementation of `verifyAudience` in this as well - or rather merge the two since one is just a special case of the other.

validator.go
```diff
@@ -255,6 +269,102 @@ func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) err
...
+			return errorIfFalse(false, ErrTokenInvalidAudience)
+		}
+
+		// Check all cmps values
+		for _, cmp := range cmps {
+			matchFound := false
+			for _, a := range aud {
+
+				// Perform constant time comparison
+				result := subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0
```
oxisto, line 312: quite a large duplicate code fragment, this is essentially what the existing `verifyAudience` does. why not use it?

validator.go
```diff
@@ -255,6 +269,102 @@ func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) err
...
+
+		// Label to break out of both loops if a match is found
+	outer:
+
+		// Check all aud values
+		for _, a := range aud {
+			for _, cmp := range cmps {
+
+				// Perform constant time comparison
+				result := subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0
```
oxisto, line 342: see above

validator.go
hdonCr: I have updated this to ` > 0`.

validator.go
hdonCr: Renamed this variable to `matchAllAud`.

hdonCr:
@oxisto, I just made some minor changes. 

Will look into the bigger changes soon (possibly merging `verifyAudience` and `verifyAudiences`, plus `expectedAud` and `expectedAuds` to some extent).

validator.go
hdonCr: I agree. I would propose to remove the validation for single audiences, since the `verifyAudiences` already supports validating a single audience by passing a slice with a single element.

I have removed the `expectedAud` field and `verifyAudiences` functions and everything related (including tests) to merge verifying one or multiple audiences. Additionally, I have added deduplication for auds and cmps and added documentation.

Note that this is a breaking change.

validator.go
```diff
@@ -255,6 +258,200 @@ func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) err
...
+
+	// If no audience is provided, return an error if required
+	if len(aud) == 0 {
+		return errorIfRequired(required, "aud")
+	}
+
+	var stringClaims string
+
+	// If matchAllAuds is true, check if all the cmps matches any of the aud
+	if matchAllAuds {
```
hdonCr, line 385: @oxisto, I am not sure whether this `verifyAudiences` function is constante time safe. Could you clarify how relevant that is? If so, I can add testing and implement this if not.

validator.go
hdonCr: See: https://github.com/golang-jwt/jwt/pull/427/files#r1948157329

validator.go
hdonCr: See my comment here: https://github.com/golang-jwt/jwt/pull/427/files#r1948157329

validator.go
hdonCr: See my comment here: https://github.com/golang-jwt/jwt/pull/427/files#r1948157329

validator.go
```diff
@@ -255,6 +258,126 @@ func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) err
...
+// Note that this does not account for any duplicate aud or cmps
+//
+// If aud is not set or an empty list, it will succeed if the claim is not required,
+// otherwise ErrTokenRequiredClaimMissing will be returned.
+//
+// Additionally, if any error occurs while retrieving the claim, e.g., when its
+// the wrong type, an ErrTokenUnverifiable error will be returned.
+func (v *Validator) verifyAudiences(claims Claims, cmps []string, required bool, matchAllAuds bool) error {
+
+	// Get the audience claim(s) from the token
```
hdonCr, line 239: @oxisto, regarding your first comment here: https://github.com/golang-jwt/jwt/pull/427/files#r1917080944

I agree that this function is quite complex. I think this is the result of the option to match all auds. Deduplicating both situations is difficult, making the function very long due to two separate flows.

Additionally, the situation in which at least one cmp must match an aud is relatively simple. However, all cmps matching all auds is more difficult to implement.

I have added documentation in an attempt to make the function more legible. If required, I can look into simplifying it further.

hdonCr:
Closed in favour of https://github.com/golang-jwt/jwt/pull/433.

PR #433, opened 2025-02-10
Implementing validation of multiple audiences

### LINKED ISSUE #342: Verifying multiple audiences

oxisto:
This PR implements a validation option to validate multiple audiences. It takes ideas from #426 and #427, but implements a sleaker and cleaner design.

Fixes #342

### DIFF
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -10,56 +10,61 @@ func TestVerifyAud(t *testing.T) {
 	var nilListInterface []interface{}
 	var intListInterface interface{} = []int{1, 2, 3}
 	type test struct {
-		Name       string
-		MapClaims  MapClaims
-		Expected   bool
-		Comparison string
-		Required   bool
+		Name        string
+		MapClaims   MapClaims
+		Expected    bool
+		Comparison  []string
+		MatchAllAud bool
+		Required    bool
 	}
 	tests := []test{
 		// Matching Claim in aud
 		// Required = true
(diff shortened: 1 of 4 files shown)


oxisto:
@mfridman

mfridman (approved):

OrkhanAlikhanov:
Can we have this released?

OrkhanAlikhanov:
Run into a bug with it and there is already a PR fixing it: https://github.com/golang-jwt/jwt/pull/441

PR #441, opened 2025-05-27
Fix bug in validation of multiple audiences

sfinnman-cotn:
In a situation where multiple audiences are validated by the validator, the order of evaluation of the for-range loop affects the result.

If we produce matches such as:

```
{
  "example.org": true,
  "example.com": false,
}
```

and we configured the validator to expect a single match on audience, the code would either:

1. produce "token has invalid audience" if "example.org" was evaluated first

2. produce a passing result if "example.com" was evaluated first

This commit fixes this bug, and adds a suite of tests as well as regression tests to prevent this issue in future.


oxisto (approved):
Thanks for spotting this. This indeed seemed to be a problem. It's really a piece of code that is not easily readable. I wonder if this could be approached from a different angle alltogether. I wonder whether there are any function available in the `slices` package that could help here.

sfinnman-cotn:
There might be! I didn't want to change the implementation too much, since i assume there is a good reason for using a map key lookup as a string comparison function. I don't want to accidentally introduce new bugs with this fix. Ill add the comment as asked for!

validator.go
```diff
@@ -235,46 +236,31 @@ func (v *Validator) verifyAudience(claims Claims, cmp []string, expectAllAud boo
...
+	if !expectAllAud {
+		for _, a := range aud {
+			// If we only expect one match, we can stop early if we find a match
+			if slices.Contains(cmp, a) {
+				return nil
+			}
 		}
 
-		stringClaims = stringClaims + a
+		return ErrTokenInvalidAudience
```
mfridman, line 252: Should we be respecting `required` bool here? I.e., we best-effort (`expectAllAud: false`) looped through the audience(s) and didn't find a single match, but if `required: false` should we still be returning an error?

sfinnman-cotn:
What more needs to be done for this to get merged @mfridman @oxisto?

mfridman:
@oxisto I'm okay we merge this, and we follow up with any improvements, if any. The only thing I can see is `if !required` and returning no error on line 252.

oxisto:
> @oxisto I'm okay we merge this, and we follow up with any improvements, if any. The only thing I can see is `if !required` and returning no error on line 252.

I am currently swamped at work and I need some time to think about this unhinged, I will have a look at this at the weekend, the rest looks good to merge.

validator.go
oxisto: The required parameter should only check if the field is set or not and we do seem to check this above when expected is not set. If expected is set we actually do not care abut required anymore, since now we can check expected vs. actual. Expected is always non-empty at this point, so if the field would empty, we would have an error here since the empty actual is not matching the expected, so we should be fine I think

oxisto: Actually this is still messy. I am thinking of removing the `required` parameter altogether. We are always setting it to true, but in theory some internal function (or a unit test) could call it with required = false, which is a mess. In the validator we already make sure we only call verifyAudience when expected auds are there and then we require them. I am replicating this check inside `verifyAudience` to make sure and set the required at this point.

@mfridman @sfinnman-cotn Could you have another look at the modified code please?

I am also thinking whether we should completely get rid of the `if len(v.expectedAud) > 0 {` check inside `Validate` as well. It's strange that the function needs to behave in a certain way with expected input. Because if you call the function with an empty expected audience (in practice this will never happen), it will fail if the claim has also an empty audience.

OrkhanAlikhanov:
Can we please have this released?

oxisto:
https://github.com/golang-jwt/jwt/releases/tag/v5.2.3

PR #443, opened 2025-06-03
Fixing bug in iat validation when enabled

the-wagness:
I was using parser options to set `WithIssuedAt()` in my application code, and noticed my unit test unexpectedly did _not_ fail when my test JWTs omitted this field.

Upon inspecting the code, I found `Validate()` seemed to pass `v.verifyIssuedAt` the value `false` for the `required` argument, which effectively ignores validation for this field when its explicitly configured via `WithIssuedAt()`.

Further, I added a testcase to the `validator_test.go` unit tests, and found that my unit test was not failing when I changed back to the old code. Upon inspection, I found that an errant if statement was only testing if the error messages matched _if_ there was an error at all, which will not test if there is _not_ an error even when one is expected.

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -117,7 +117,7 @@ func (v *Validator) Validate(claims Claims) error {
 
 	// Check issued-at if the option is enabled
 	if v.verifyIat {
-		if err = v.verifyIssuedAt(claims, now, false); err != nil {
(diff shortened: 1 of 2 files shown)


the-wagness:
Closing this PR as I have found this behavior is intentional per: https://github.com/golang-jwt/jwt/issues/411

PR #445, opened 2025-06-07
replaced interface{} to any in KeyFunc

aachex:
Hello, I replaced interface{} to its alias any in jwt.KeyFunc
<br>
```go
type Keyfunc func(*Token) (interface{}, error) // before
type Keyfunc func(*Token) (any, error)         // after
```

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -22,29 +22,29 @@ var (
 	jwtTestEC256PublicKey  crypto.PublicKey
 	jwtTestEC256PrivateKey crypto.PrivateKey
 	paddedKey              crypto.PublicKey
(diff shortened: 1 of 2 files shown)


mfridman:
If you want to go all the way, you can run this command against the root of the repository:

```sh
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix ./...
```

mfridman (approved):

mfridman:
And then look for all comments `interface{}` and update them to `any` for completeness. Careful with a crude search/replace, please check those updates make sense.

oxisto (approved):

Camaster52:
Cool! The guy has a great future.

PR #452, opened 2025-07-23
Bump Go version to indicate correct minimum requirement

oxisto:
The minimum requirement that we have was silently bumped to 1.21 in https://github.com/golang-jwt/jwt/pull/441 because of the `slices` package. It seems that we did not update the `go.mod` when we updated our CI range, because CI did not fail because it was not testing older versions.

We probably should just update the `go.mod` when we update the CI target in the future? Although we could theoretically stay at 1.21 in terms of the code base.


oxisto:
> What's up with the go1.22 failures, do we have some old build tags we can get rid of now?

https://github.com/golang-jwt/jwt/blame/3839817bf313f2cfe58853dcbe542ffd06300831/rsa_pss.go#L1-L2

I think we can just remove those two build tags. They were probably from a time when Go 1.4 was new. But I think we still have some other build tags for the errors, don't we?

mfridman:
> I think we can just remove those two build tags. They were probably from a time when Go 1.4 was new. But I think we still have some other build tags for the errors, don't we?

Yep, that's what I was thinking. Remove those _very old_ build tags entirely.

oxisto:
> > I think we can just remove those two build tags. They were probably from a time when Go 1.4 was new. But I think we still have some other build tags for the errors, don't we?
> 
> Yep, that's what I was thinking. Remove those _very old_ build tags entirely.

The error handling ones are Go 1.20 build tags, which technically we would also not need anymore when we bump the minimum version to Go 1.21. I am not a BIG fan of doing this now within a major version but we (I) accidentically already released the latest v5 version which needs 1.21 without bumping it so I guess the damage is done already :(

mfridman:
It's probably defensible to bump the minimum Go version at some point. We probably don't need to keep up with the latest 2 versions like the Go team does.

We do call out that we keep up with the supported versions. And we rarely (if ever) pull in new features. Basically, I think it's fine as it is for now. Any concerns?

oxisto:
> It's probably defensible to bump the minimum Go version at some point. We probably don't need to keep up with the latest 2 versions like the Go team does.

Agree

> 
> We do call out that we keep up with the [supported versions](https://github.com/golang-jwt/jwt?tab=readme-ov-file#supported-go-versions). And we rarely (if ever) pull in new features. Basically, I think it's fine as it is for now. Any concerns?

100 % agree, no concerns from my side then.

> 
> The alternative is to remove `slices` usage and release a version that's compatible with go1.18? But tbh I'm not overly bothered by bumping it up to go1.21.

No, let's go forward, not backwards. I'll clean it up and remove the remaining code for Go < 1.21.

PR #457, opened 2025-08-01
Additional test for CustomClaims that validates unmarshalling behaviour

### LINKED ISSUE #453: custom claim with an integer field fails parsing
Hello,
if you define a custom claim like this:
```
type myClaims struct {
	jwt.RegisteredClaims
	MyField int `json:"myField"`
}
````
then all parsing methods accepting claim definitions seem to fail with a JSON unmarshal error. As far as I can tell this is because the `golang-jwt` package interprets all numbers as float64.

`golang-jwt` version: v5.2.2

### PR DESCRIPTION
**Pull request by equalsgibson**
Closes #453 

Adds an additional test to parser_test.go, that validates the library can parse the valid JWT Claims in the token to a custom Go struct with different types - validates: 

-    a string
-    a number
-    an object (JSON object)
-    an array
-    a boolean
-    null

Which should provide coverage for all valid data types as per the [JSON spec](https://www.json.org/json-en.html)

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -132,6 +132,26 @@ var jwtTestData = []struct {
 		nil,
 		jwt.SigningMethodRS256,
 	},
+	/*
+		custom claims with JSON Data Types
+		Payload:
+		{
+		  "stringClaim": "string",
+		  "intClaim": -1,
+		  "boolClaim": true,
+		  "uintClaim": 1,
+		  "floatClaim": 5.01,
+		  "sliceClaim": [
+		    -1,
+		    0,
+		    1
+		  ],
+		  "objectClaim": {
+		    "property": "something"
+		  },
+		  "nilPointerClaim": null
+		}
+	*/
 	{
 		"custom claims with JSON Data Types",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: parser_test.go [hunk 1]
```diff
@@ -113,6 +132,27 @@ var jwtTestData = []struct {
 		nil,
 		jwt.SigningMethodRS256,
 	},
+	{
+		"custom claims with JSON Data Types",
```
oxisto, line 136: Can you add a comment to this with the JSON payload?

```
{
  "stringClaim": "string",
  "intClaim": -1,
  "boolClaim": true,
  "uintClaim": 1,
  "floatClaim": 5.01,
  "sliceClaim": [
    -1,
    0,
    1
  ],
  "objectClaim": {
    "property": "something"
  },
  "nilPointerClaim": null
}
```

that's helpful in debugging. We might add that for the remaining structures in the future as well.

oxisto (approved):
Good idea. One small nit. It does however not really increase our test coverage, since this is basically handled by `encoding/json`. However, there were some issues creeping up where folks didn't get their custom claims parsed, so this should help to weed out any errors here.

parser_test.go
equalsgibson: Added the comment above the testdata struct, so it is easy to see the expected input and output

equalsgibson:
Hi @mfridman, the reasoning for adding the tests is that there is not documented proof for custom claims being marshalled currently. I agree that the std lib does this already - but as there is a wrapper function around the use of json.unmarshal, (with the parse functions), this ensures that if any changes are made to the implementation in this package, that current functionality (custom claims types) would not break.

Happy to close this out if not needed or if you feel it is not warranted 👍

mfridman (approved):

PR #459, opened 2025-08-02
338 - Update godoc example func to reference New() rather than NewWithClaims()

### LINKED ISSUE #338: Token.New example is for the wrong func
The example for Token.New() actually demonstrates Token.NewWithClaims(). It should be demonstrating Token.New().

It additionally demonstrates signing, which I guess is ok but it seems out of scope, given that the signing example can just be part of the documentation for that method ...

equalsgibson:
Closes #338

### DIFF
--- a/hmac_example_test.go
+++ b/hmac_example_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"time"
 
 	"github.com/golang-jwt/jwt/v5"
 )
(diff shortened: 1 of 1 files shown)


hmac_example_test.go
```diff
@@ -25,18 +24,17 @@ func init() {
 
 // Example creating, signing, and encoding a JWT token using the HMAC signing method
 func ExampleNew_hmac() {
-	// Create a new token object, specifying signing method and the claims
-	// you would like it to contain.
-	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
-		"foo": "bar",
-		"nbf": time.Date(2015, 10, 10, 12, 0, 0, 0, time.UTC).Unix(),
-	})
+	// Create a new token object, specifying signing method
```
oxisto, line 27: Actually I think it would make more sense to just rename the example function instead. In this example, `NewWithClaims` should be the preferred way to create the token. In the majority of the cases you want to add custom claims to your token, so this should be the "default" example.

oxisto (changes requested):

mfridman (approved):

PR #473, opened 2025-09-22
Add TPM signature reference

salrashid123:
Adds an example reference for TPM based signatures

i can also add in pkcs-11 reference included here https://github.com/salrashid123/golang-jwt-pkcs11


oxisto (approved):
That's very cool!

mfridman:
No objection to adding this one. But we should probably figure out a baseline for adding these.

Maybe this should live in the docs or a separate "community" repository?

PR #485, opened 2025-11-22
Feat: MapClaims Key Type-Safety

YlanzinhoY:
### Before
image

### After
image

---------
### I can blend the new way with the old way and it works.

image

image

---------

### ctrl+space

image

### DIFF
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -192,7 +192,7 @@ func signToken() error {
 	// add command line claims
 	if len(flagClaims) > 0 {
 		for k, v := range flagClaims {
-			claims[k] = v
+			claims[jwt.ClaimsType(k)] = v
 		}
 	}
 
--- a/cmd/jwt/main.go
+++ b/cmd/jwt/main.go
@@ -192,7 +192,7 @@ func signToken() error {
 	// add command line claims
 	if len(flagClaims) > 0 {
 		for k, v := range flagClaims {
-			claims[jwt.ClaimsType(k)] = v
+			claims[jwt.MapClaimsKey(k)] = v
 		}
 	}
 
--- a/map_claims.go
+++ b/map_claims.go
@@ -5,19 +5,31 @@ import (
 	"fmt"
 )
 
+// ClaimsType is a type alias for claim keys in the [MapClaims] type.
 type ClaimsType string
 
 var (
+	// Exp is the "exp" claim key according to RFC 7519 section 4.1.4.
 	Exp ClaimsType = "exp"
+
+	// Nbf is the "nbf" claim key according to RFC 7519 section 4.1.5.
 	Nbf ClaimsType = "nbf"
+
+	// Iat is the "iat" claim key according to RFC 7519 section 4.1.6.
 	Iat ClaimsType = "iat"
+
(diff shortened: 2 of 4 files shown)


oxisto:
I had another look at this and I really like the concept. 

@mfridman what are your thoughts about merging this into v5? It should not really break anything except in case where people are doing really really really weird things such as manually casting a `map[string]any` into `MapClaims` (or vice-versa).

I am not really sure about the `ClaimsType` name, should probably be clearer that it only relates to `MapClaims`.

YlanzinhoY:
> I had another look at this and I really like the concept.
> 
> @mfridman what are your thoughts about merging this into v5? It should not really break anything except in case where people are doing really really really weird things such as manually casting a `map[string]any` into `MapClaims` (or vice-versa).
> 
> I am not really sure about the `ClaimsType` name, should probably be clearer that it only relates to `MapClaims`.

Which name do you think is best?  I can make a change if you want.

oxisto:
> ClaimsType

Maybe `MapClaimsType`

mfridman:
> I had another look at this and I really like the concept.
> 
> @mfridman what are your thoughts about merging this into v5? It should not really break anything except in case where people are doing really really really weird things such as manually casting a `map[string]any` into `MapClaims` (or vice-versa).
> 
> I am not really sure about the `ClaimsType` name, should probably be clearer that it only relates to `MapClaims`.

I don't think we can take this chance, there are too many dependents. And from a maintainer's perspective, I am always biased towards extremely strong backwards compatibility.

oxisto:
> > I had another look at this and I really like the concept.
> > @mfridman what are your thoughts about merging this into v5? It should not really break anything except in case where people are doing really really really weird things such as manually casting a `map[string]any` into `MapClaims` (or vice-versa).
> > I am not really sure about the `ClaimsType` name, should probably be clearer that it only relates to `MapClaims`.
> 
> I don't think we can take this chance, there are too many dependents. And from a maintainer's perspective, I am always biased towards extremely strong backwards compatibility.

Understandable. Since "main" is now the to-be-v6, we can still merge it, after we agree on a name for key-type.

oxisto (approved):
I settled on `MapClaimsKey` for the type name. @mfridman your turn

mfridman:
> Understandable. Since "main" is now the to-be-v6, we can still merge it, after we agree on a name for key-type.

Probably outside the scope of this PR, maybe we can chat on Slack/email. 

But wdyt about keeping `main` as `v5` and having a long-lived `v6.0.0-rc.1` (or `v6.0.0-beta.1`) branch with all the breaking changes? Keep in mind we'll need to change the go.mod, so `/v5` users shouldn't be affected and we can continue to add bug and security fixes on main (rebasing the feature branch as needed).

oxisto:
> > Understandable. Since "main" is now the to-be-v6, we can still merge it, after we agree on a name for key-type.
> 
> Probably outside the scope of this PR, maybe we can chat on Slack/email.
> 
> But wdyt about keeping `main` as `v5` and having a long-lived `v6.0.0-rc.1` (or `v6.0.0-beta.1`) branch with all the breaking changes? Keep in mind we'll need to change the go.mod, so `/v5` users shouldn't be affected and we can continue to add bug and security fixes on main (rebasing the feature branch as needed).

-> mail ;)

Issue #238, opened 2022-09-15
modify header claims


afidalgo:
Hello, im trying jwt-go and found no way to modify header claims. Is there a way that im missing or i need to modify the code to accept it?

If you dont mind i can provide the diff after the change.

thanks in advance, BR

afidalgo:
i made the change of this issue opening text and an extra modification, once verified, get the payload claims and also the header claims. Please let me know if you want the diff or the PR, thanks again

afidalgo:
sorry, my mistake on last note, deleted to avoid confusion.

mfridman:
What is the header you're referring to, any chance you can share some code snippet or link so we can better understand?

The smallest reproducible example would help us here..

afidalgo:
for sure. im working with something called stir and shaken (specs: atis100082, atis100074, atis 100085 and the rfc's referenced from those documents, 8224, 8225,8226,7519, and others)
One thing stir and shaken does is to use the ability to extend the passport token (a jwt) and this extension (at least part of it) is done in the Header.
This is an example of this header for non-diverted call scenarios (atis100082)

`
{ "alg": "ES256", "typ": "passport", "ppt": "shaken", "x5u": "https://cert.example.org/passport.crt"}
`

for diverted/re routed calls (atis100085) header is

`
{
"alg":"ES256","ppt":"div","typ":"passport","x5u":"https://cert.example2.net/passport.cer"
}
`
So had the need to put extra headers manually, i modified the token.go to create a new method to allow this (maybe is not the better way but it works :) )

`
func NewWithClaimsAndHeaders(method SigningMethod, claims Claims, headers map[string]interface{}) *Token {
	customHeader := make(map[string]interface{})
	customHeader["typ"] = "JWT"
	customHeader["alg"] = method.Alg()

	for k, v := range headers {
		if k != "alg" {
			customHeader[k] = v
		}
	}
	return &Token{
		Header: customHeader,
		Claims: claims,
		Method: method,
	}
}
`

BR

oxisto:
You can just use `jwt.NewWithClaims` to retrieve a `Token` object and just modify the `Header` of it.

You can see an example here https://github.com/oxisto/oauth2go/blob/b135c891f36ea062cb72a9f8d34a7900e0bf3783/server_test.go#L61-L66

afidalgo:
great!! thanks, that helped.

PR #498, opened 2026-03-07
feat: Add helpers for setting `typ` and `cty` header parameters

### LINKED ISSUE #497: Add helpers for setting `typ` and `cty` header parameters

Example:
```go
    token.Header["typ"] = "JWT"
```
```go
    token.SetType("at+jwt")
    token.SetContentType("JWT")
```

Arabasta:

Example:
```go
    token.Header["typ"] = "JWT"
```
This PR adds helper methods:
```go
    token.SetType("at+jwt")
    token.SetContentType("JWT")
```

These modify Token.Header internally and do not change signing or validation behavior.

This improves discoverability for these registered header parameters.

Happy to adjust naming or placement if a different approach is preferred.

Closes https://github.com/golang-jwt/jwt/issues/497

### DIFF
--- a/token.go
+++ b/token.go
@@ -100,3 +100,30 @@ func (t *Token) SigningString() (string, error) {
 func (*Token) EncodeSegment(seg []byte) string {
 	return base64.RawURLEncoding.EncodeToString(seg)
 }
+
+// SetType sets the "typ" (type) header parameter.
+// According to https://datatracker.ietf.org/doc/html/rfc7519#section-5.1,
+// this parameter is optional.
+// Passing an empty string removes the parameter.
+func (t *Token) SetType(typ string) {
+	t.setHeaderParam("typ", typ)
+}
+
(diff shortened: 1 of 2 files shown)


oxisto:
Ok seems there is some demand here.. There are literally 3 PRs (or it's just AI bait for easy issues). It is some un-canonical Go to have `SetXXX` functions and a simple `token.Header["typ"]=value` seems not to hard to do. I wonder if we actually pre-allocate the `Header` properly and maybe it makes sense to have const values for `typ` and other well-known headers. Any opinions on that @mfridman?

mfridman:
Indeed, I suspect it's easy pickings. I've seen plenty of Go libraries with setters (and getters) so probably not uncommon.

I don't have a strong preference here, but either a constant or a setter is fine IMO.

oxisto:
Went with constants. see #535

PR #503, opened 2026-04-05
Fix ParseUnverified returning error on malformed signature

### LINKED ISSUE #499: `5.3.1` now treats malformed signatures as an error in `ParseUnverified`, which does not align with the documented behavior
`5.3.1` now treats malformed signatures as an error [here](https://github.com/golang-jwt/jwt/pull/414/changes#diff-83eb8e32639d01cf443d6d8bde24c1c8be78766090d8c5f8586c36250cfedca6R198) in `ParseUnverified` which differs from the documented behavior of that method which before ignored signature parsing issues, matching the documented behavior of **not** validating the signature. This was pushed as a patch to `5.3.1` but feels more like a breaking change that should be published separately, as it changes client expectations for how signatures are treated.

Bahtya:
## Problem

Since v5.3.1, `ParseUnverified` returns `ErrTokenMalformed` when base64 decoding the signature fails. This contradicts the documented behavior of not validating the signature.

Callers using `ParseUnverified` to inspect tokens without validation now get unexpected errors on tokens with malformed signatures.

## Solution

Ignore the `DecodeSegment` error for the signature portion in `ParseUnverified`, since the method explicitly should not care about signature validity.

Fixes #499

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -193,10 +193,8 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 	}
 
 	// Parse token signature
-	token.Signature, err = p.DecodeSegment(parts[2])
-	if err != nil {
-		return token, parts, newError("could not base64 decode signature", ErrTokenMalformed, err)
-	}
+	token.Signature, _ = p.DecodeSegment(parts[2])
+	// Ignore error: ParseUnverified does not validate the signature per documentation
 
 	return token, parts, nil
 }


oxisto:
Thanks for your contribution. This is somewhat of a duplicate of https://github.com/golang-jwt/jwt/pull/500.

So my *personal* opinion of that: Yes it was not good that we (accidentally) introduced something which was not 100 % backwards compatible from a behavior part of view - while we stayed backwards compatible in the API.

BUT, in my opinion it DOES NOT contradict the function documentation, since that only says that it does ignore the *validation* of the signer - which is still 100 % the case. The only thing it does check is whether the token itself is malformed, in terms of how the standard dictates it should look like. If you are running into this error your token is definitely either corrupt or you are doing something which is not standard compliant and since this library aims to be standard compliant, your use case is probably not catered by this library.

Personally, I would have not have this function *at all*, because I think all use-cases involving this design is inherently dangerous. If one is not verifying the token (because it is supposedly already verified), why transferring the token at all? Just parse and validate the token in one place and forward the necessary extracted information instead of a token.

Bahtya:
Thanks for the thoughtful response @oxisto. I appreciate your perspective.

You make a fair point — the function documentation does describe ignoring *validation* of the signer, and the malformed token check is consistent with standard compliance. I can see the argument that this is expected behavior rather than a bug.

Regarding the duplicate with #500 — understood. If the maintainers decide this is not something the library should support, I'm happy to close this PR. I respect the project's direction.

Thanks again for taking the time to explain the rationale.

PR #505, opened 2026-04-14
fix: treat exp=0 as expired (epoch time) instead of unset

### LINKED ISSUE #496: token is not treated as expired when exp = 0
Discovered that when using `Parse` with no options, a token with the claim `"exp": 0` will not produce an `ErrTokenExpired` error as expected.

Did some tinkering and used the `WithJSONNumber()` parser option, and the same token with 0 expiration was treated as expired.

They should be treated the same regardless of which type they are parsed to because the content of the token is the same.

Yanhu007:
Fixes #496

## Problem

`MapClaims.parseNumericDate` has an inconsistency between `float64` and `json.Number` handling for the value `0`:

| Parse mode | `"exp": 0` result | Token expired? |
|---|---|---|
| Default (float64) | `nil` (claim not set) | No ❌ |
| WithJSONNumber() | epoch time (1970-01-01) | Yes ✅ |

Per RFC 7519, `exp: 0` means "expires at Unix epoch (1970-01-01)", which is always in the past.

## Fix

Change the `float64(0)` case to return `newNumericDateFromSeconds(0)` instead of `nil`, matching the `json.Number` behavior.

```go
// Before
case float64:
    if exp == 0 { return nil, nil }  // treated as "not set"

// After
case float64:
    if exp == 0 { return newNumericDateFromSeconds(0), nil }  // treated as epoch
```

All existing tests pass.

### DIFF
--- a/map_claims.go
+++ b/map_claims.go
@@ -51,7 +51,7 @@ func (m MapClaims) parseNumericDate(key string) (*NumericDate, error) {
 	switch exp := v.(type) {
 	case float64:
 		if exp == 0 {
-			return nil, nil
+			return newNumericDateFromSeconds(0), nil
 		}
 
 		return newNumericDateFromSeconds(exp), nil


map_claims.go
```diff
@@ -51,7 +51,7 @@ func (m MapClaims) parseNumericDate(key string) (*NumericDate, error) {
 	switch exp := v.(type) {
 	case float64:
 		if exp == 0 {
```
oxisto, line 53: So I think that this line here was actually there to catch `"exp": ""`, which we should treat as not set instead of `0` but it if course also runs into this corner case when the value is actually 0. We could probably change this if to something like `key == ""` because the "else" branch should then already cover the 0 part.

oxisto (approved):

oxisto:
Seems like https://github.com/golang-jwt/jwt/pull/509 is the up-do-date PR tackling this problem

PR #509, opened 2026-05-16
mapclaims: stop treating exp=0 as a missing claim

### LINKED ISSUE #496: token is not treated as expired when exp = 0

c-tonneslan:
Fixes #496.

`MapClaims.parseNumericDate` short-circuits and returns `nil` whenever a `float64` value is exactly 0. That made `\"exp\": 0` (a token that expired at the unix epoch) invisible to `verifyExpiresAt`, which checks for `exp == nil` and concludes the claim isn't set, so the token was accepted as valid.

The `json.Number` branch a few lines below had no such carve-out. Parsing the same payload via `WithJSONNumber()` came back correctly as expired, which is how the reporter discovered the inconsistency in the first place.

Removed the special case so both code paths agree: an `exp` of 0 returns a `NumericDate(0)` and the validator surfaces `ErrTokenExpired`.

Same logic applies to `nbf` and `iat` parsed from the same helper. `nbf=0` keeps meaning \"valid since the epoch\" (no behavior change in practice). `iat=0` now reports as 1970-01-01, which is what the spec says it is.

Added `TestMapClaims_GetExpirationTime_ZeroIsExpired` covering both the `float64` and `json.Number` paths.

### DIFF
--- a/map_claims.go
+++ b/map_claims.go
@@ -50,10 +50,6 @@ func (m MapClaims) parseNumericDate(key string) (*NumericDate, error) {
 
 	switch exp := v.(type) {
 	case float64:
-		if exp == 0 {
-			return nil, nil
-		}
-
 		return newNumericDateFromSeconds(exp), nil
 	case json.Number:
 		v, _ := exp.Float64()
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -216,3 +216,23 @@ func TestMapClaims_GetExpirationTime_ZeroIsExpired(t *testing.T) {
 		})
 	}
 }
+
+// A string exp must come back as ErrInvalidType, not as a stealth
+// "claim not present" via the old float64==0 shortcut. Empty string is
+// the case worth pinning down explicitly.
+func TestMapClaims_GetExpirationTime_StringIsInvalidType(t *testing.T) {
+	for name, claims := range map[string]MapClaims{
+		"empty string": {"exp": ""},
+		"non-empty":    {"exp": "foo"},
+	} {
+		t.Run(name, func(t *testing.T) {
+			_, err := claims.GetExpirationTime()
+			if err == nil {
+				t.Fatalf("expected an error, got nil")
+			}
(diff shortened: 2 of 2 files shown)


oxisto:
Can you consider my comment in the other PR https://github.com/golang-jwt/jwt/pull/505#discussion_r3078882909

c-tonneslan:
Yeah, took a look. My PR removes the `if` entirely rather than rewording its condition, so the `float64` case falls through to `newNumericDateFromSeconds(exp)` (returns `NumericDate(0)` for 0, matching the `json.Number` branch).

For what it's worth I don't think `"exp": ""` actually reaches `case float64`, it decodes as `string` and hits the default `ErrInvalidType`. But happy to defer if you'd rather take the #505 approach.

oxisto:
> Yeah, took a look. My PR removes the `if` entirely rather than rewording its condition, so the `float64` case falls through to `newNumericDateFromSeconds(exp)` (returns `NumericDate(0)` for 0, matching the `json.Number` branch).
> 
> For what it's worth I don't think `"exp": ""` actually reaches `case float64`, it decodes as `string` and hits the default `ErrInvalidType`. But happy to defer if you'd rather take the #505 approach.

Ah. Good point. Can we make a unit test about this special case and then I think its good to go.

c-tonneslan:
Done, pushed a test covering empty-string and non-empty-string `exp` values, both come back as `ErrInvalidType`.

oxisto (approved):

mfridman:
Ouch, this will likely impact _someone_ relying on the `nil` return to detect "no exp present". But I think it's defensible as it's a bug fix.

mfridman (approved):

PR #510, opened 2026-05-21
fix: extended token expiry message with duration

TarasHots:
Prior to v5, the expiry validation error included the duration by which the token had expired (e.g. token is expired by 7m0s). 

This was removed in v5, losing useful debugging information.

This PR restores that behavior — the error message now includes the elapsed duration since expiry, making it easier to diagnose issues like clock skew or misconfigured TTLs without having to manually compare timestamps.

Before: `token is expired`
After:  `token is expired by 7m0s`

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -182,7 +182,7 @@ func (v *Validator) verifyExpiresAt(claims Claims, cmp time.Time, required bool)
 		return errorIfRequired(required, "exp")
 	}
 
-	return errorIfFalse(cmp.Before((exp.Time).Add(+v.leeway)), ErrTokenExpired)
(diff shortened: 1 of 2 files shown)


mfridman:
Nit, with leeway=10m and a token whose exp was 15m ago, the error reads "expired by 5m0s."

For debugging clock skew or a misconfigured TTL, the raw delta from exp is probably the more informative number (the leeway is already known config)?

Maybe `cmp.Sub(exp.Time)` is sufficient. Figured I'd mention it.

TarasHots:
@mfridman applied, thanks! Left the unary "+" as is since it was already there. I know it does not make a difference here, but e.g. `verifyIssuedAt` or `verifyNotBefore` have "-". Just sticking to the same codestyle

oxisto (approved):

mfridman (approved):

PR #511, opened 2026-05-24
fix: return ErrInvalidType for an invalid aud claim type in MapClaims

ultramcu:
## Problem

`MapClaims.parseClaimsString` (used by `MapClaims.GetAudience`) silently returns an empty audience and a `nil` error when the `aud` claim has an unexpected type such as a number, boolean or object:

```go
aud, err := jwt.MapClaims{"aud": 123}.GetAudience()
// before: aud == nil, err == nil   (silently swallowed)
```

This is inconsistent with the rest of the library:

- `parseString` (`iss`, `sub`) returns `ErrInvalidType` for a wrong type.
- `parseNumericDate` (`exp`, `nbf`, `iat`) returns `ErrInvalidType` for a wrong type.
- `parseClaimsString` itself already returns `ErrInvalidType` when an **element** of a `[]any` audience is not a string.
- `ClaimStrings.UnmarshalJSON` (the `RegisteredClaims` path) returns `ErrInvalidType` for the same input.

So a malformed `aud` scalar was the only claim that failed silently. The practical effect is a confusing diagnostic: with `WithAudience(...)` the validator reports `"aud is required"` (because the audience came back empty) instead of surfacing the actual invalid type. The token is still rejected, so this is not a security issue — it's a correctness/consistency fix that makes malformed tokens easier to debug.

## Fix

Add a `default` case to the type switch that returns `ErrInvalidType`, plus an explicit `nil` case so that an **absent or null** `aud` keeps meaning "no audience" (the claim is optional — no behaviour change there).

## Tests

Added `TestMapClaims_GetAudience`, a table test documenting the full contract (missing/null → no error; string/`[]string`/`[]any` → parsed; wrong element / wrong scalar type → `ErrInvalidType`). It fails before the change on the number/bool/object cases and passes after.

`go test -race ./...`, `go vet ./...` and `gofmt` are clean. No existing test changed.

### DIFF
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -3,6 +3,7 @@ package jwt
 import (
 	"encoding/json"
 	"errors"
+	"reflect"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


map_claims_test.go
```diff
@@ -192,3 +194,45 @@ func TestMapClaims_parseString(t *testing.T) {
 		})
 	}
 }
+
+func TestMapClaims_GetAudience(t *testing.T) {
+	tests := []struct {
+		name    string
+		m       MapClaims
+		want    ClaimStrings
+		wantErr bool
```
oxisto: Can we assert the specific error type here please?

oxisto (approved):
Makes sense, small nit on the test error but overall a good change.

ultramcu:
Thanks for the review @oxisto! Addressed the nit: the test now carries the expected error per case (`wantErr error`, set to `ErrInvalidType` for the invalid-type cases) and asserts with `errors.Is(err, tt.wantErr)` instead of a generic bool flag.

oxisto:
@mfridman

mfridman (approved):

PR #513, opened 2026-05-25
feat: add WithIssuedAtRequired parser option

wyf027:
## Summary

Fixes #489  
Refs #411

Maintainers previously agreed not to change `WithIssuedAt()` semantics in v5 (validate only when present), and instead add a dedicated required option like `WithExpirationRequired` / `WithNotBeforeRequired`.

This PR adds:

- `requireIat` field on `Validator`
- `WithIssuedAtRequired()` parser option (enables iat verification and makes the claim mandatory)
- `Validate()` now passes `v.requireIat` to `verifyIssuedAt`
- Tests covering optional vs required iat behavior

Usage for callers who need a mandatory `iat` claim:

```go
jwt.Parse(tokenString, keyfunc,
    jwt.WithIssuedAtRequired(),
    jwt.WithAudience(expectedAudience),
)
```

## Test plan

- [x] `go test ./...`

### DIFF
--- a/parser_option.go
+++ b/parser_option.go
@@ -48,11 +48,21 @@ func WithTimeFunc(f func() time.Time) ParserOption {
 	}
 }
 
-// WithIssuedAt returns the ParserOption to enable verification
-// of issued-at.
+// WithIssuedAt returns the ParserOption to enable verification of issued-at
+// when the claim is present.
 func WithIssuedAt() ParserOption {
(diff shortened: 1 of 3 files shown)


oxisto:
I'm hesitant to add new parser options at this stage unless there is a HUGE demand for it, since we potentially want to re-work the validation system in v6. The current version with all of these option functions is not very flexible :(

PR #514, opened 2026-05-25
fix(parser): ignore malformed signature in ParseUnverified

### LINKED ISSUE #499: `5.3.1` now treats malformed signatures as an error in `ParseUnverified`, which does not align with the documented behavior

wyf027:
## Summary

- `ParseUnverified` no longer fails when the signature segment cannot be base64url-decoded
- Malformed signature bytes are treated as absent signature data, matching the method's documented contract of not validating signatures
- Restores behavior expected by callers that only need claims from tokens whose signature segment is corrupted or intentionally ignored

## Problem

Since v5.3.1, `ParseUnverified` returns `ErrTokenMalformed` when the signature segment is invalid base64. That is stricter than the documented behavior ("does not validate the signature") and breaks callers that previously could still read claims from such tokens.

## Test plan

- [x] Added `TestParser_ParseUnverifiedIgnoresMalformedSignature`
- [x] `go test ./...`

Fixes #499

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -192,11 +192,9 @@ func (p *Parser) ParseUnverified(tokenString string, claims Claims) (token *Toke
 		return token, parts, newError("signing method (alg) is unspecified", ErrTokenUnverifiable)
 	}
 
-	// Parse token signature
-	token.Signature, err = p.DecodeSegment(parts[2])
-	if err != nil {
-		return token, parts, newError("could not base64 decode signature", ErrTokenMalformed, err)
-	}
+	// Parse token signature. Signature validation is intentionally skipped in
+	// ParseUnverified; a malformed signature segment must not fail parsing.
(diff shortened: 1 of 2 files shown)


oxisto:
Duplicate of https://github.com/golang-jwt/jwt/pull/500 and potentially not even an issue (see https://github.com/golang-jwt/jwt/issues/499#issuecomment-4325975737)

PR #515, opened 2026-05-26
fix: return ErrInvalidType for an invalid aud claim type in MapClaims

wyf027:
## Summary

Fixes #511

`MapClaims.parseClaimsString` silently returned an empty audience for scalar `aud` values with unexpected types (number, bool, object). This aligns `aud` parsing with `iss`/`sub`/`exp` by returning `ErrInvalidType`, while preserving missing/null `aud` as optional.

## Test plan

- [x] Added `TestMapClaims_GetAudience` table test
- [x] `go test ./...`

Made with [Cursor](https://cursor.com)


oxisto:
Please don't open pull requests for something which already has a PR which is partially reviewed and accepted.

PR #517, opened 2026-06-11
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets covering critical code paths. Part of OSS-Fuzz integration proposal (google/oss-fuzz#15663). All targets verified: go test -fuzz=. -fuzztime=30s

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,38 @@
+name: CIFuzz
+on:
+  pull_request:
+    paths:
+      - '**'
+  push:
+    branches: [main, master]
+permissions:
+  contents: read
+jobs:
+  fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address]
+    steps:
+    - name: Build Fuzzers (${{ matrix.sanitizer }})
+      id: build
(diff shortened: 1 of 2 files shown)


fuzz_test.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2026 Google LLC
```
oxisto: We dont usually include a separate copyright header and we are using MIT license, not Apache 2.0

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,43 @@
+name: CIFuzz
+on:
+  pull_request:
+    paths:
+      - '**.go'
+      - '.github/workflows/cifuzz.yml'
+  push:
+    branches: [main, master]
```
oxisto: we dont have master

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,43 @@
...
+
+permissions:
+  contents: read
+  security-events: write
```
oxisto: what does this entail actually?

oxisto (changes requested):

canolgun:
@oxisto Thank you for the review! All three issues fixed:

1. **Copyright header**: Changed to MIT ()
2. **Branch**: Removed , only  now
3. **Permissions**: Simplified to just  — the SARIF upload step handles its own permissions

Ready for re-review.

PR #520, opened 2026-07-01
feat: add required issued-at validation option

happysnaker:
## Summary
- add `WithIssuedAtRequired()` to require the `iat` claim while also enabling issued-at validation
- thread the new `requireIat` validator setting through claim validation
- add regression tests for missing required `iat` claims

## Why
Issue #489 points out that `WithIssuedAt()` validates future `iat` values but still leaves the claim optional, unlike the existing required-claim options such as `WithExpirationRequired()` and `WithNotBeforeRequired()`.

This change keeps the current `WithIssuedAt()` behavior intact for callers that only want the sanity check, while providing an explicit option for callers that want `iat` to be mandatory.

Closes #489

### DIFF
--- a/map_claims_test.go
+++ b/map_claims_test.go
@@ -90,6 +90,15 @@ func TestMapclaimsVerifyIssuedAtInvalidTypeString(t *testing.T) {
 	}
 }
 
+func TestMapclaimsVerifyIssuedAtRequired(t *testing.T) {
+	mapClaims := MapClaims{}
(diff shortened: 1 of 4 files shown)


happysnaker:
Quick follow-up: I pushed a tiny formatting-only commit after checking the failed build logs. The actual failure was just `gofmt` on `validator_test.go`; there was no functional test failure in the change itself.

PR #533, opened 2026-08-22
Fixes #413: set token.Signature in ParseUnverified

### LINKED ISSUE #413: `ParseUnverified` doesn't set the token `Signature` field

TheRodzz:
Fixes #413. This ensures the `Signature` field is correctly populated on the token when calling `ParseUnverified`, similar to the behavior of `Parse`. I have also updated the relevant test.


TheRodzz:
Hi  @oxisto , could you please review this, Thanks!

mfridman:
Any reason you can't upgrade to /v5 which if iirc has the signature set?

PR #534, opened 2026-08-28
feat(token): add SetType and SetContentType header helpers to Token (fix #497)

### LINKED ISSUE #497: Add helpers for setting `typ` and `cty` header parameters

Jaymr1031-Z:
Fixes #497

### Summary
Adds convenience helper methods on Token to set standard RFC 7519 header parameters typ (Type, Section 5.1) and cty (Content Type, Section 5.2) without requiring manual map mutation.

### Changes
- Added Token.SetType(typ string)
- Added Token.SetContentType(cty string)
- Added TestToken_SetTypeAndContentType in token_test.go


oxisto:
Close as duplicate of https://github.com/golang-jwt/jwt/pull/498

